PR #31, opened 2026-04-24
feat(auth): users/sessions/devices schema + auth db layer (PR1/4)

seamus-sloan:
## Summary

PR1 of a 4-PR stack implementing F0.3 auth ([docs/roadmap/0-3-auth.md](docs/roadmap/0-3-auth.md)). This PR is pure data layer — zero user-facing change, no routes gated. Safe to merge and sit idle.

- **Migration `0004_auth.sql`** — `users`, `devices`, `sessions`, `secrets` tables; explicit permission booleans (not a role bitmask, per Calibre-inspection #9); `registration_enabled` setting flipped to `0` atomically after the first user. All time columns are Unix seconds for cheap integer compares.
- **`db/src/auth.rs`** — Argon2id password hashing at OWASP 2024 floor (19 MiB, t=2, p=1), PHC string storage so params can rotate on verify; 256-bit CSPRNG session tokens stored as `SHA-256(token)` so a DB leak cannot be replayed as live auth; race-free first-user-admin via `BEGIN IMMEDIATE`; timing-equalized unknown-user login path (runs argon2 verify against a sentinel hash); 5-strike per-account lockout with 15-min cooldown; session + device CRUD; session signing-key persistence.
- **21 unit tests**, including the concurrent-registration race, expired/revoked session rejection, and an explicit "raw token is never persisted" assertion against the `sessions` table.
- Deps added to `db/Cargo.toml`: `argon2`, `rand`, `sha2`, `base64`, `thiserror`.

## Test plan

- [x] \`cargo test -p omnibus-db auth::\` — 21/21 green
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [x] \`cargo fmt --check\` — clean
- [ ] Review: confirm the \`BEGIN IMMEDIATE\` + row-count pattern is the right race-free shape (alternatives: unique sentinel row, partial unique index — both more awkward in SQLite)
- [ ] Review: confirm Argon2id parameters (19 MiB, t=2, p=1) are acceptable on the slowest target hardware
- [ ] Review: confirm the common-password reject list (50 entries) is enough for v1.0

## Stack

1. **PR1 (this)** — db foundation
2. PR2 — server plumbing: extractors, \`/api/login\`, \`/api/logout\`, \`/api/register\`, CSRF, rate-limits, tower-sessions wiring, \`OMNIBUS_INITIAL_ADMIN\` boot hook
3. PR3 — flip the switch: gate existing routes + ship \`login.rs\` / \`register.rs\` pages + Playwright \`auth.spec.ts\`
4. PR4 — mobile bearer flow

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,6 +48,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "argon2"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
+dependencies = [
+ "base64ct",
+ "blake2",
+ "cpufeatures",
+ "password-hash",
+]
+
+[[package]]
 name = "askama_escape"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -268,6 +280,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
 dependencies = [
  "serde_core",
+]
+
+[[package]]
+name = "blake2"
+version = "0.10.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "46502ad458c9a52b69d4d4d32775c788b7a1b85e8bc9d482d92250fc0e3f8efe"
+dependencies = [
+ "digest",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


seamus-sloan:
Addressed all 6 review comments in a9a66ad1:

1. **Lockout reset bug** — `verify_login` now treats `failed_login_count` as 0 once `locked_until` has elapsed, so a single post-cooldown failure no longer instantly re-locks. Added regression test `login_lockout_resets_after_cooldown_elapses`.
2. **Doc wording** — module docstring now says "exact SHA-256 hash match" instead of "constant-time hash compare".
3. **Unknown session kind** — migration now has `CHECK(kind IN ('cookie', 'bearer'))`; `lookup_session` returns `SessionNotFound` on unexpected values instead of defaulting to Cookie.
4. **Migration idempotency** — `INSERT INTO settings ('registration_enabled', '1') ON CONFLICT(key) DO NOTHING`.
5. **Redundant index** — removed `idx_sessions_token_hash`; the `UNIQUE` constraint already provides the index.
6. **Unicode-aware length** — `validate_password` now uses `password.chars().count()` with a comment explaining why.

Tests: 22/22 green (one added). Clippy + fmt clean.

PR #58, opened 2026-05-09
fix(frontend): style password and email inputs on auth pages

seamus-sloan:
## Summary
- Extend `.settings-field` input selector to cover `input[type="password"]` and `input[type="email"]` so the login and register password (and register email) fields pick up the same dark-themed background, border, and focus ring as the settings page text inputs. Previously they fell through to browser defaults and rendered as bare white boxes.

## Test plan
- [ ] Visit `/login` — password field has dark background, subtle border, matches settings page styling.
- [ ] Visit `/register` — both email and password fields are styled.
- [ ] Focus state on each: outline removed, border switches to `#3b82f6`.
- [ ] Settings page text inputs unchanged (regression check).

## Notes
- CSS-only change inside the `STYLES` const in `frontend/src/lib.rs`. No markup changes; auth.rs already wraps inputs in `.settings-field`.

### DIFF
--- a/frontend/src/lib.rs
+++ b/frontend/src/lib.rs
@@ -323,7 +323,10 @@ h1 { font-size: 1.4rem; margin-bottom: 0.5rem; }
   font-weight: 500;
   color: #cbd5e1;
 }
-.settings-field input[type="text"], .settings-input {
+.settings-field input[type="text"],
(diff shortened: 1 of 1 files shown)


seamus-sloan:
@copilot review

PR #59, opened 2026-05-09
fix(frontend): anchor body gradient with background-attachment: fixed

seamus-sloan:
## Summary
- Add `background-attachment: fixed` to the `body` rule so the radial gradient stays anchored to the viewport instead of stretching with document height. On long pages (e.g. landing with many books) the unfixed gradient washed out toward the pale top color (`#1f2937`) for nearly the entire scroll, killing the intended dark fade.

## Test plan
- [ ] On `/` with enough books to scroll several viewports, the background looks the same dark tone at the bottom as at the top.
- [ ] No visible "jump" or seam when scrolling — gradient stays put under the content.
- [ ] Short pages (`/settings`, `/login`) look unchanged.
- [ ] iOS / Android mobile shells unaffected (rule is on `body`; native targets don't render the body gradient anyway).

## Notes
- Single-line CSS change in the `STYLES` const in `frontend/src/lib.rs`.


seamus-sloan:
@copilot review

PR #184, opened 2026-05-25
ci: add cargo-audit security job (#173)

### LINKED ISSUE #173: [Tech Debt] Medium - No cargo audit job in CI; security advisories go undetected
Neither `.github/workflows/rust.yml` nor `e2e.yml` runs `cargo audit` (or `cargo deny`) against the dependency tree. Security advisories published to the RustSec advisory database — covering transitive dependencies like `reqwest`, `sqlx`, `axum`, `tokio`, `argon2`, and `rustls` — will not surface as CI failures. A contributor adding a dependency with a known CVE, or a downstream advisory published against an already-used crate, passes CI silently. Omnibus already uses several security-sensitive crates: Argon2id for password hashing, rustls for outbound TLS to Open Library, tower-http for cookie-authed HTTP, and axum for all user-facing endpoints.

This is **Medium** severity. Omnibus is self-hosted and single-tenant, which limits exposure compared to a SaaS product. But the auth surface (session tokens, password hashes) and the fact that cover/photo uploads accept arbitrary binary multipart payloads make dependency hygiene important. Advisory scanning is a zero-cost gate once wired in — `cargo audit` runs in seconds against the `Cargo.lock` without a full build.

Sources: `.github/workflows/rust.yml` (current jobs: `fmt`, `check` — no audit step). The `Cargo.lock` is checked in (correctly), so `cargo audit` can run against the exact pinned versions without a compile.

Add a `security` job to `rust.yml` that runs `cargo audit --deny warnings` (or `cargo deny check advisories`). It should trigger on the same conditions as the `check` job (PRs and `main` push). Wire it through `nix develop --command cargo audit` to stay inside the Nix environment already established by the `check` job. If `cargo deny` is preferred for broader control (license checks, duplicate detection), add a `deny.toml` to the repo root. Either way, the job should block merge on any RUSTSEC advisory against the current `Cargo.lock`.

seamus-sloan:
## Summary
- Add a `security` job to `.github/workflows/rust.yml` that runs `cargo audit --deny warnings` inside the Nix dev shell. RustSec advisories against transitive deps (reqwest, sqlx, axum, tokio, argon2, rustls, tower-http) were previously never surfaced as CI failures; this job blocks merge on any advisory.
- The new job is modeled on the existing `check` job: same `actions/checkout@v4` + `DeterminateSystems/nix-installer-action@v22` + `nix-community/cache-nix-action@v7` steps and the same `contents: read` / `id-token: write` / `actions: write` permissions. It uses a distinct `nix-audit-` cache prefix so it doesn't clobber the `check` job's `nix-rust-` cache.
- It inherits the workflow's existing triggers (`pull_request` with paths incl. `flake.nix` + `.github/workflows/rust.yml`, and `push` to `main`) — no `on:` change needed.
- Wire `pkgs.cargo-audit` into `devShells.default` in `flake.nix` so the command is available both in CI and locally.
- Chose `cargo audit` over `cargo deny` (less config, matches the issue's primary suggestion); no `deny.toml` added.

Closes #173

## Test plan
- [ ] CI `cargo audit` job runs on this PR (it touches `flake.nix` and `.github/workflows/rust.yml`, both in the trigger paths) and passes / surfaces any advisory.
- [ ] Confirm the `security` job appears as a required/visible check and blocks on a non-zero `cargo audit` exit.

## Notes
- This PR edits a CI workflow (`.github/workflows/rust.yml`) and modifies `flake.nix`.
- `cargo audit` was NOT run locally: the agent sandbox has no Nix installed (`nix` not on PATH, no `/nix` store), so `nix develop --command cargo audit` could not be exercised. CI will be the first run. YAML was validated as well-formed.
- `Cargo.lock` was not changed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UvoQ8RDugg1XJbfRdmKJYD)_

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -106,6 +106,8 @@ jobs:
       # Required by nix-community/cache-nix-action `purge: true` so it can
       # delete its own expired cache entries via the Actions cache API.
       actions: write
+      # Required to post / update the sticky audit-report comment on PRs.
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


seamus-sloan:
The `cargo audit` job did its job on the first run — it surfaced real advisories in the dependency tree. Pushed `eb3dde9` to make the gate green-able while keeping it meaningful. Decisions (flagging for your review):

**Fixed**
- `rustls-webpki 0.103.12` → `0.103.13` (RUSTSEC-2026-0104, reachable panic in CRL parsing). Clean one-line `Cargo.lock` bump — **calling out the lockfile change**.

**Dropped `--deny warnings`** (deviation from #173's suggestion). The Dioxus desktop/mobile stack drags in the EOL `gtk-rs` GTK3 bindings plus a few other unmaintained/unsound transitive crates (`glib`, `proc-macro-error`, `fxhash`, `rand`) — **14 warnings with no available fix**. Denying warnings would pin the gate permanently red with nothing actionable. Without the flag, the gate still fails on any *vulnerability* advisory, which is the security-critical part.

**Ignored `RUSTSEC-2023-0071`** (`rsa` 0.9.10, Marvin timing attack). No fixed release exists, and `rsa` is reached **only via `sqlx-mysql`**, which this SQLite-only app never uses at runtime. Rationale documented inline in the workflow.

If you'd prefer a stricter posture, two alternatives: (a) keep `--deny warnings` and enumerate all ~15 current advisories in an `audit.toml` so only *new* ones fail (more upkeep), or (b) trim `sqlx` to sqlite-only features so `rsa` (and the postgres `rand` path) drop out of the tree entirely — a bigger dependency change I held off on. Happy to do either; let me know.

---

seamus-sloan:
Restructured the security job in d14e138d so it does what you asked: cargo audit runs on every PR, and findings get posted as a sticky PR comment instead of failing the gate. No audit.toml allow-list (A), no sqlx trimming (B).

Mechanism:
- `cargo audit --json` runs with `continue-on-error: true`, writing `audit.json`. The previous `--ignore RUSTSEC-2023-0071` flag is gone — rsa will show up in the report alongside anything else.
- A formatter step turns `audit.json` into a markdown table (advisory ID, package@version, title, URL) and posts/updates a single sticky comment per PR via `marocchino/sticky-pull-request-comment` keyed on a `<!-- cargo-audit-report -->` header.
- The job stays green even when advisories are found — the comment is the signal, not the check status.
- Handles the "cargo-audit itself failed to run" case (e.g. advisory DB fetch failure) by reporting that explicitly in the comment rather than silently claiming "no findings".
- `audit.json` is uploaded as a 14-day artifact for any deeper inspection.
- Added `pull-requests: write` to the job so it can post comments.

This PR's own next run should produce the first sticky comment listing the current advisories (rsa, plus whatever else is now in the tree). The follow-on PR comment will overwrite that in-place.

---

PR #293, opened 2026-05-30
refactor: prune unpopulated EbookMetadata wire fields (#174)

### LINKED ISSUE #174: [Tech Debt] Low - EbookMetadata wire type carries ~11 fields always None/0 from every DB read
`shared/src/lib.rs` defines `EbookMetadata` with fields for Dublin Core metadata collected during EPUB parsing: `rights`, `source`, `coverage`, `dc_type`, `dc_format`, `relation`, `epub_version`, `resource_count`, `spine_count`, `toc_count`, and `contributors`. All of these fields are always returned as `None` or `0` from every DB read path (`list_books`, `get_book`, `search_books`, `get_author`, `get_series`). The normalized schema in `0002_normalized_schema.sql` does not have columns for these fields — they were extracted from EPUBs by `db/src/ebook.rs` during parsing into the in-memory `IndexedBook` struct but were never persisted to the DB. Every `EbookMetadata` returned over the wire (to web, mobile, or Dioxus RSX) carries these dead fields serialized as `null`/`0`, adding payload bloat and creating a contract that implies the data is available when it isn't.

This is **Low** severity for correctness — no code currently reads these fields to make decisions, so the dead data doesn't cause bugs. It is **Medium** severity for maintainability: any contributor who sees `EbookMetadata.resource_count` and tries to use it in UI will get `0` for every book and have no hint that the field is unpopulated. The `contributors` field (distinct from `creators`) is particularly misleading because the `shared/src/lib.rs` type definition and the `Contributor` struct are fully fleshed out, implying the data is available.

Sources: `shared/src/lib.rs` — `EbookMetadata` struct definition. `db/src/queries.rs` — `row_to_ebook` and the `list_books`/`get_book` projection blocks; search for `rights`, `source`, `coverage`, `epub_version`, `resource_count`, `spine_count`, `toc_count`, `contributors` — none appear in any SQL projection. `db/src/ebook.rs` — `IndexedBook` correctly populates these fields during parsing but they are never written to the DB.

Two options: (a) Add schema columns for the missing fields and populate them in the `replace_books` insert path — the data is already extracted by `ebook.rs` into `IndexedBook`, so it's a matter of threading it into the SQL INSERT and adding the corresponding columns to the SELECT projections; (b) Remove the dead fields from `EbookMetadata` and `shared/src/lib.rs` if they are not planned for the F1.x or F2.x roadmap — removing them simplifies the wire type, eliminates payload bloat, and stops misleading contributors. Needs human judgment on which roadmap features depend on these fields.

seamus-sloan:
## Summary
- Removed eleven dead fields from `shared::EbookMetadata` that every DB read path returned as `None` / `0` / `[]`: `rights`, `source`, `coverage`, `dc_type`, `dc_format`, `relation`, `epub_version`, `resource_count`, `spine_count`, `toc_count`, `contributors`. They were parsed from EPUBs by `db::ebook::parse` but never threaded into the schema, so the wire type carried bloat that misled contributors into thinking the data was available.
- Merged `<dc:contributor>` entries into `creators` at parse time (creators first, contributors in OPF source order). The schema already stored both in `books_authors_link` and the read paths already surfaced contributors as `creators`, so this just makes the parser output match the read-path output. `insert_author_links` and `insert_fts_row` drop their contributors-chain branch — the names are reached via `m.creators` instead.
- Simplified `book_detail::kicker_label` after removing `dc_type` (its sole consumer, which always took the `None` branch). Kicker now renders `"Book"` or `"Book · YEAR"`.

Closes #174.

## Audit trail per field

| Field | Consumer audit | Decision |
|---|---|---|
| `rights`, `source`, `coverage`, `dc_format`, `relation` | `rg --type rust` → only `shared/src/lib.rs` and the four `db/src/books.rs` SELECT initializers. No reader. | Remove. |
| `dc_type` | Read by `frontend/src/pages/book_detail.rs::kicker_label`, but DB always returns `None`, so the reader always hit the `None` arm. Simplified the kicker. | Remove + simplify reader. |
| `epub_version`, `resource_count`, `spine_count`, `toc_count` | Parsed only; no reader anywhere. | Remove. |
| `contributors` | Wire field always empty on read (schema doesn't distinguish creator/contributor in `books_authors_link`). Parser populated it; `db::sync::insert_author_links` and `insert_fts_row` read it during insert and merged it into the same join table as `creators`. | Merge into `creators` at parse time; drop the field. |
| `Contributor` struct | Still backs the live `creators` field across DB, server, frontend. | Keep. |

Roadmap grep (`docs/roadmap/`) found no references to any of the removed field names; only adjacent uses of `coverage` and `relation` as English nouns in unrelated prose.

## Notes
- Test contract update: `db/tests/public_domain_epubs.rs` asserted the parser-stage shape. Seven fixtures have `<dc:contributor>` entries (Sedgwick on `civics_and_health.epub`, Barrie on `charles_frohman.epub`, six translators on `gogol_dramatische_werke.epub`, Merrill on `little_women.epub`, Morley on `mariucha.epub`, Beets on `woordenlijst_nederlandsche_taal.epub`, Hetherington on `works_of_george_gillespie.epub`) — each is now appended to the expected `authors` array, with an inline comment pointing at this issue.
- The four `kicker_label` unit tests in `book_detail.rs` collapsed into two covering the simplified function.
- No new migration: the schema never carried these columns. No `Cargo.lock` changes.
- This is technically a wire-protocol break: an older mobile client that deserializes `EbookMetadata` strictly (no `serde(default)`) and depends on the dead fields being present would break. Given the fields were always `null`/`0`/`[]`, no caller could have been making a decision on them, but flagging for awareness.

## Test plan
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy --all-targets -- -D warnings` clean (default-members; per CLAUDE.md, `--all-features` hits the pre-existing `omnibus-frontend` mobile-feature conflict unrelated to this change).
- [x] `cargo test` (default-members): 247 passing, 0 failing.
- [x] `cargo test -p omnibus-db`: 330 passing across unit + fixture + public-domain integration suites.
- [x] `cargo test -p omnibus-frontend --features server`: 99 passing.
- [x] `cargo test -p omnibus`: 128 passing.

https://claude.ai/code/session_01V1khPHj86bq18g4aVfbeuC

---
_Generated by [Claude Code](https://claude.ai/code/session_01V1khPHj86bq18g4aVfbeuC)_


seamus-sloan:
**Playwright CI failure — diagnosed as flake; recommend re-running the job.**

Failing run: https://github.com/seamus-sloan/omnibus/actions/runs/26680661597/job/78640352730.

This PR has two behavior changes that could *in principle* trip Playwright — `kicker_label` simplification and the `<dc:contributor>` → `creators` merge — but both check out:

**Kicker.** Confirmed `db/src/books.rs:185, 331, 584, 902` on `main` already hard-codes `dc_type: None` in every `row_to_ebook` / projection callsite. The simplified `kicker_label(year)` is observationally identical to the old `kicker_label(None, year)`. Independently: `rg "kicker" ui_tests/playwright/tests/` returns nothing — no spec asserts on kicker text in the first place.

**Contributor merge.** `ui_tests/playwright/tools/make_epub.ts:403, 425–427` only emits `<dc:creator>` — no `<dc:contributor>` for any synthetic fixture. The seven public-domain EPUBs that carry contributors already had their merged-author lists committed in `ui_tests/playwright/tests/fixtures/epubs.ts` on `main` (commit `667a24e`, *fix(playwright): align contributor fixtures with indexer + raise seed budget*). `expectRowMatches` in `tests/utils/ebooks.ts:70` does an exact `toHaveText(authors.join(", "))` against those values, which already match the merged shape on `main`. `book_detail.spec.ts:52` only asserts `toContainText(TARGET.authors[0])` (first author), so even row-shape shifts wouldn't break it.

**Seed count.** `FIXTURE_BOOKS.length` (49) matches `28 generated + 21 public_domain` on disk — aligned.

**No diff to `ui_tests/playwright/` at all in this PR.**

Combined with PR #291's identical flake-only Playwright failure: two CSS / DB-only PRs both flaking on Playwright at the same time strongly suggests an environmental issue (seed timing, server boot race, browser launch). Recommend re-running the `playwright` job from the Actions UI on both PRs.

---

PR #463, opened 2026-06-11
fix: add stable key: to creator byline loop in book_detail (#402)

### LINKED ISSUE #402: [Tech Debt] Medium - Five Dioxus list renderings missing key= attribute (new instances after #379 and #231)
## Summary
Five `for` loops inside `rsx!` blocks render dynamic lists without a `key:` attribute. Issues #379 and #231 (both CLOSED) addressed earlier instances; these five are new or were missed. Without `key:`, Dioxus uses position-based reconciliation: adding, removing, or reordering a list item causes every subsequent item to re-render unnecessarily, and stateful components can receive the wrong identity after an update.

## Severity: Medium
Position-based diffing is a correctness issue on any list that can change (reordered, filtered, grown, shrunk). The series book list, the speed/sleep preset buttons, and the format badge spans are all rendered from dynamic data and can change in length or order at runtime.

## Evidence
- `frontend/src/pages/series.rs:124` — `for book in s.books.iter() { article { ... } }` — book cards rendered without `key:`; books can be reordered by sort key
- `frontend/src/pages/listen/speed_panel.rs:51` — `for val in PRESETS.iter() { button { ... } }` — speed preset buttons without `key:`
- `frontend/src/pages/listen/sleep_panel.rs:37` — `for preset in PRESETS.iter() { button { ... } }` — sleep timer preset buttons without `key:`
- `frontend/src/pages/landing/table.rs:426` — `for fmt in display_book.formats.iter() { span { ... } }` — format badge spans without `key:`; format list length varies per book
- `frontend/src/pages/landing/filters.rs:172` — `for (key, count) in counts.into_iter() { ... }` — filter facet items without `key:`; facet counts change as filters are applied

Rule reference: `.claude/rules/04-playwright.md` — rendered lists must carry stable `key:` props to prevent position-based reconciliation bugs.

## Acceptance criteria
- [ ] All five `for` loops in the Evidence list have a `key:` attribute on the top-level child element that is stable and unique within the list.
- [ ] `cargo clippy -p omnibus-frontend --features server --all-targets -- -D warnings` passes.
- [ ] `cargo test -p omnibus-frontend --features server` passes unchanged.
- [ ] Where Playwright E2E specs cover the affected pages, they pass without regression.

## Suggested approach
Add `key:` to the outermost element in each loop, using the most stable natural identifier available:

```rust
// Before (series.rs)
for book in s.books.iter() {
    article { ... }
}

// After
for book in s.books.iter() {
    article { key: "{book.uuid}", ... }
}
```

```rust
// Before (speed_panel.rs, sleep_panel.rs) — static presets
for val in PRESETS.iter() {
    button { ... }
}

// After — preset value is stable
for val in PRESETS.iter() {
    button { key: "{val}", ... }
}
```

```rust
// Before (filters.rs)
for (key, count) in counts.into_iter() {
    li { ... }
}

// After — key variable is already in scope
for (key, count) in counts.into_iter() {
    li { key: "{key}", ... }
}
```

## Notes for the implementer
- For `landing/table.rs:426`, the format string itself (e.g. `"epub"`, `"m4b"`) is the stable key — a book can only have one entry per format type.
- For `series.rs:124`, use `book.uuid` not the loop index — index-based keys defeat the purpose of keyed reconciliation.
- The `key` variable in `filters.rs:172` is already named `key` in the destructuring — this fix is a one-word addition.

seamus-sloan:
## Summary

- PR #410 already addressed five of the six list-loop sites flagged in #402; the issue was reopened against the creator byline in `frontend/src/pages/book_detail.rs` per a follow-up comment.
- Add `key: "{i}-{creator.name}"` to both the `Link` (author present) and `span` (author missing) branches of the `for (i, creator) in b.creators.iter().enumerate()` loop under `BdTitleCol`. Dioxus now reconciles by stable identity instead of position when creators are added/removed/reordered (e.g. via a metadata override).
- The literal `", "` separator between creators stays unkeyed — text nodes don't need keys.

## Notes vs. the issue runbook

- Sites 1-5 (`series.rs`, `listen/{speed,sleep}_panel.rs`, `landing/table.rs`, `landing/filters.rs`) already carry `key:` attributes on `main` from PR #410, so no changes in those files in this PR.
- The runbook also suggested renaming the destructured `key` in `landing/filters.rs` to `facet` to avoid shadowing the `key:` rsx attribute name. PR #410's body called that site a "Sniffer false positive" — the existing `key: "{key}"` interpolation works correctly because `key:` is parsed as an rsx attribute, not a Rust binding. Leaving that file untouched to keep this PR scoped to the reopen reason; the rename can be a separate cleanup if desired.
- `series.rs` uses `book.id` rather than the issue's suggested `book.uuid` because `EbookMetadata` has `id: i64` (always present) and `unique_identifier: Option<String>` (nullable). That call was made in PR #410 and remains correct.

## Test plan

- [x] `cargo clippy -p omnibus-frontend --features server --all-targets -- -D warnings` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes
- [x] `cargo test -p omnibus-frontend --features server` — 149 passed, 0 failed
- [x] `cargo fmt` — no changes

This touches markup contracts (no role/label/testid changes, only a virtual-DOM hint), so it should not require new Playwright coverage — but `run_ui_tests` is set on the PR as a safety check.

Closes #402

---
_Generated by [Claude Code](https://claude.ai/code/session_01BSYobgzHvj4g9bnYoFQVdx)_

### DIFF
--- a/frontend/src/pages/book_detail.rs
+++ b/frontend/src/pages/book_detail.rs
@@ -465,14 +465,14 @@ fn BdTitleCol(
                         if i > 0 { ", " }
                         if let Some(author_id) = creator.id {
                             Link {
-                                key: "{i}-{creator.name}",
+                                key: "id-{author_id}",
                                 to: Route::AuthorDetail { id: author_id },
                                 class: "bd-author-link",
                                 "{creator.name}"
                             }
                         } else {
                             span {
-                                key: "{i}-{creator.name}",
+                                key: "name-{creator.name}-{creator.role:?}-{creator.file_as:?}",
                                 class: "bd-author-link",
                                 "{creator.name}"
                             }
--- a/frontend/src/pages/book_detail.rs
+++ b/frontend/src/pages/book_detail.rs
@@ -464,9 +464,18 @@ fn BdTitleCol(
                     for (i, creator) in b.creators.iter().enumerate() {
                         if i > 0 { ", " }
                         if let Some(author_id) = creator.id {
-                            Link { to: Route::AuthorDetail { id: author_id }, class: "bd-author-link", "{creator.name}" }
+                            Link {
+                                key: "{i}-{creator.name}",
+                                to: Route::AuthorDetail { id: author_id },
+                                class: "bd-author-link",
+                                "{creator.name}"
+                            }
                         } else {
-                            span { class: "bd-author-link", "{creator.name}" }
+                            span {
+                                key: "{i}-{creator.name}",
+                                class: "bd-author-link",
+                                "{creator.name}"
+                            }
                         }
                     }
                 }


seamus-sloan:
Addressed both Copilot review comments — keys now use the stable creator identity instead of the loop index:

- **id-present branch** (`creator.id == Some(author_id)`): `key: "id-{author_id}"` — the canonical `authors.id` primary key, which is the most stable handle Dioxus can reconcile against across edits/reorders.
- **id-missing branch** (`creator.id == None`): `key: "name-{creator.name}-{creator.role:?}-{creator.file_as:?}"` — combines all three stable per-contributor fields on `shared::ebook::metadata::Contributor` (`name: String`, `role: Option<String>`, `file_as: Option<String>`) to minimise collisions when the same name appears with different roles (e.g. author vs. translator).

Kept `.enumerate()` because the `if i > 0 { ", " }` text-node separator still uses `i`; only the key derivation changes.

Local: `cargo fmt`, `cargo clippy -p omnibus-frontend --features server --all-targets -- -D warnings`, and `cargo test -p omnibus-frontend --features server` (149 passed) all clean.

---

PR #674, opened 2026-07-02
fix: enforce SESSION_BATCH_CAP on rpc_record_sessions

seamus-sloan:
## Summary
- Hoist the session-batch cap (`500`) out of `server::backend::progress` and into `omnibus_shared` as `pub const SESSION_BATCH_CAP: usize = 500;`, with a `///` comment on why it's compile-time (coordinated server + client bump, not a per-deploy tweak).
- `server::backend::progress::post_sessions` now imports `SESSION_BATCH_CAP` from `omnibus_shared` and its 422 message references the new name.
- `frontend::rpc::rpc_record_sessions` (the web RPC path that previously lacked a batch cap) now short-circuits with a `ServerFnError` before any per-record validation or DB work, matching the tone of the REST 422. The check lives in a small server-gated `check_session_batch_size` helper (mirroring the existing `validate_author_photo_url` pattern) so a unit test can exercise the over-cap rejection without spinning up the dioxus server-fn router.

## Test plan
- [ ] `just lint` (`cargo fmt --check` + clippy across the crate/feature matrix)
- [ ] `cargo test -p omnibus-shared` — exercises the shared constant via existing `SessionReport` tests
- [ ] `cargo test -p omnibus post_sessions_rejects_oversized_batch` — 422 path on the mobile REST route
- [ ] `cargo test -p omnibus-frontend --features server check_session_batch_size` — over-cap and at-cap paths for the web RPC

- The 422 message wording (`"batch too large: N records exceeds maximum of CAP"`) is kept verbatim across REST and RPC so log-scraping and client-side error surfacing stay consistent.
- The REST-side test constant was renamed from `MAX_SESSION_BATCH` to `SESSION_BATCH_CAP` in the same commit; nothing else referenced the old name.
- Deliberately kept the RPC handler on non-transactional inserts (the pre-existing behavior noted in the doc comment) — that's a separate defect from the cap gap, out of scope for this PR.

Closes #609

---
_Generated by [Claude Code](https://claude.ai/code/session_015XGkb1sHd7fp3UR5VC9Vkw)_

---
### Adversarial review

**Verdict: approved-with-notes** — resolves #609 cleanly; only cosmetic nits.

- Acceptance criteria met: `SESSION_BATCH_CAP` is defined once in `omnibus_shared::progress` (flat re-exported via `pub use progress::*` from `shared/src/lib.rs`), imported by both `server::backend::progress` and `frontend::rpc` without copy-paste; the `check_session_batch_size` guard runs as the **first** line of `rpc_record_sessions`'s body, so it fires strictly before the per-item `validate()` loop and any DB work. `MAX_SESSION_BATCH` grep on the branch returns zero hits — the rename is complete.
- Scope is tight: only the four cited files (`shared/src/progress.rs`, `server/src/backend/progress.rs`, `server/src/backend/progress/tests.rs`, `frontend/src/rpc.rs`) touched, no `Cargo.lock` churn, no migrations.
- 422 wording is kept verbatim across REST and RPC (`"batch too large: N records exceeds maximum of CAP"`), so log-scraping stays consistent. The REST handler still uses `>` (not `>=`), and the RPC helper matches — the new `check_session_batch_size_accepts_at_cap` test pins that boundary, avoiding an off-by-one divergence between the two clients.
- Style nit (non-blocking): `check_session_batch_size` returns `std::result::Result<(), ServerFnError>`, while the sibling `validate_author_photo_url` uses the shorter `Result<&str, ServerFnError>` alias from `dioxus::prelude`. Both compile; consistent style would drop the `std::result::` prefix.
- Style nit (non-blocking): the new `#[cfg(feature = "server")] use omnibus_shared::SESSION_BATCH_CAP;` line was inserted between `AUTHOR_PHOTO_URL_MAX_LEN` and `BookSuggestion`, which breaks strict alphabetical order (A → S → B). The surrounding block is grouped by "server-gated single symbol imports" more than by alpha order, so this is a very minor smell.
- Out-of-scope callout is honest: the RPC path's non-transactional insert loop (a DB error mid-batch commits already-persisted rows and loses the count) is documented in the updated doc comment and explicitly deferred — that's a separate defect from the cap gap and belongs in its own change.

### DIFF
--- a/frontend/src/rpc.rs
+++ b/frontend/src/rpc.rs
@@ -34,6 +34,12 @@ use omnibus_shared::{
 #[cfg(feature = "server")]
 use omnibus_shared::AUTHOR_PHOTO_URL_MAX_LEN;
 
+// `check_session_batch_size` is server-gated (called only from the server-side
+// body of `rpc_record_sessions`), so keep the import `server`-gated too — the
+// client build would otherwise flag it unused.
+#[cfg(feature = "server")]
+use omnibus_shared::SESSION_BATCH_CAP;
+
 // `BookSuggestion` is only constructed in the server-side `rpc_get_suggestions`
 // body; gate it so the web/mobile client builds don't flag an unused import.
 #[cfg(feature = "server")]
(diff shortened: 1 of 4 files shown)


seamus-sloan:
Superseded by #679, which merged the same `SESSION_BATCH_CAP` fix for #609 into `main` (commit 16073256) — #609 is now closed as completed. Closing this as a duplicate.

PR #678, opened 2026-07-02
fix: cap session-report batches on rpc_record_sessions (#609)

seamus-sloan:
## Summary
- Hoist `MAX_SESSION_BATCH` from `server::backend::progress` into `omnibus_shared` as `pub const SESSION_BATCH_CAP: usize = 500` so both the REST twin and the web RPC reference one source of truth.
- Add the same length check to `frontend::rpc::rpc_record_sessions` that `post_sessions` has been enforcing — an authenticated web caller can no longer smuggle an unbounded batch into the sequential write loop and pin the SQLite WAL write lock.
- Extract `validate_session_batch(&[SessionReport]) -> Result<(), ServerFnError>` so the length + per-item checks can be exercised directly from `#[cfg(all(test, feature = "server"))]` tests (mirrors the pre-existing `validate_author_photo_url` pattern — the fullstack server-fn body can't be reached without spinning up the whole stack).

## Files touched
- `shared/src/progress.rs` — add `SESSION_BATCH_CAP` with `///` cap docs.
- `server/src/backend/progress.rs` — replace the local `pub(super) const MAX_SESSION_BATCH` with the shared name; update the `post_sessions` docstring to point at the web twin.
- `server/src/backend/progress/tests.rs` — import the shared name, rename the reference in `post_sessions_rejects_oversized_batch`.
- `frontend/src/rpc.rs` — `server`-gated `use omnibus_shared::SESSION_BATCH_CAP` + `validate_session_batch`; `rpc_record_sessions` now short-circuits via the helper; three new tests in the `mod tests` block (over-cap rejection, at-cap accept, per-item validator still fires under the cap).

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings` (default members: server, shared, frontend)
- [x] `cargo test -p omnibus-shared`
- [x] `cargo test -p omnibus-db`
- [x] `cargo test -p omnibus`
- [x] `cargo test -p omnibus-frontend --features server` — three new `validate_session_batch_*` tests pass, existing `post_sessions_rejects_oversized_batch` still green.

## Notes
- Skipped `cargo clippy -p omnibus-mobile` and `cargo clippy --all-features` in this sandbox: mobile needs GTK-3 that isn't available here, and `--all-features` on `omnibus-frontend` combines the mutually incompatible `mobile` + `web` features (pre-existing on `origin/main`, not introduced by this change). Neither is touched by this fix.
- `SESSION_BATCH_CAP` stays a compile-time constant — no `.env.example` change needed, per the acceptance criteria on the issue.

Closes #609

---
_Generated by [Claude Code](https://claude.ai/code/session_01YMQ1M4kQQA3SmTkRCXFv61)_

### DIFF
--- a/frontend/src/rpc.rs
+++ b/frontend/src/rpc.rs
@@ -34,6 +34,11 @@ use omnibus_shared::{
 #[cfg(feature = "server")]
 use omnibus_shared::AUTHOR_PHOTO_URL_MAX_LEN;
 
+// Only `validate_session_batch` (server-gated) and its tests reference this
+// cap; same reason for the `server` gate as `AUTHOR_PHOTO_URL_MAX_LEN` above.
+#[cfg(feature = "server")]
+use omnibus_shared::SESSION_BATCH_CAP;
+
 // `BookSuggestion` is only constructed in the server-side `rpc_get_suggestions`
 // body; gate it so the web/mobile client builds don't flag an unused import.
 #[cfg(feature = "server")]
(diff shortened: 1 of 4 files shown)


seamus-sloan:
Superseded by #679, which merged the same `SESSION_BATCH_CAP` fix for #609 into `main` (commit 16073256) — #609 is now closed as completed. Closing this as a duplicate.

Note for the record: #679 landed the length-only variant (`check_session_batch_cap(&reports)` with the per-item `validate()` loop left inline). The extra regression test from this PR — `validate_session_batch_propagates_per_item_validation_error`, which guards against the cap check clobbering the per-item loop — was not carried over to main.

PR #680, opened 2026-07-02
feat: keyboard-accessible journal spoiler reveal (#644)

seamus-sloan:
## Summary
- Rendered spoiler wrapper is now a native `<button class="spoiler" type="button" aria-expanded="false">` — Tab reaches it, Enter/Space fire click without a keydown listener, and screen readers announce it as a button.
- Ammonia allowlist extended to keep exactly that shape (`class="spoiler"`, `type="button"`, `aria-expanded` ∈ {true,false}); off-list classes/types and spoofed aria values are stripped, and the checkbox-only `<input>` guard is untouched.
- Document-level click handler flips `aria-expanded` alongside `.revealed` so assistive tech reflects the toggled state.
- CSS resets UA button chrome (`font/color/padding/border/appearance`) on `button.spoiler` so it still reads as inline prose.

Files touched:
- `db/src/journals/markdown.rs` — new markup + sanitizer allowlist + updated tests (happy path renames to `wraps_spoiler_markers_in_keyboard_reachable_buttons`, plus a new `disallows_arbitrary_button_classes_and_attrs` to lock in the button allowlist per rule 03).
- `frontend/src/pages/book_detail/journal.rs` — click handler toggles `aria-expanded`; no keydown path needed since a native button already emits click on Enter/Space.
- `frontend/assets/atrium.css` — button chrome reset for `button.spoiler`.
- `ui_tests/playwright/tests/flows/journal.spec.ts` — existing mouse test now also asserts `aria-expanded` flips; new **"reveals a spoiler with Enter and toggles it back with Space"** action test drives the button via `page.keyboard.press` after `focus()` (rule 04 semantic-first selectors).
- `.claude/architecture.md` — refreshes the `db::journals::markdown` blurb citing the new emitted markup (rule 98).

## Test plan
- [x] `cargo fmt --check` (via `cargo fmt`, no diff)
- [x] `cargo clippy --all-targets` (clean)
- [x] `cargo clippy -p omnibus-frontend --features server --all-targets` (clean)
- [x] `cargo test -p omnibus-db` (passes, incl. new `wraps_spoiler_markers_in_keyboard_reachable_buttons` and `disallows_arbitrary_button_classes_and_attrs`)
- [x] `cargo test -p omnibus`
- [x] `cargo test -p omnibus-frontend --features server`
- [x] `cargo test -p omnibus-shared`
- [ ] `cd ui_tests/playwright && npx playwright test tests/flows/journal.spec.ts` — new Enter/Space keyboard test **not exercised locally** (requires the running dev stack); wired to match the existing `.spoiler` locator and existing rebuild-and-poll fixtures.

## Notes
- **Skipped `cargo clippy -p omnibus-mobile`** — the mobile crate needs GTK-3 dev headers from the `.#mobile` nix shell (pre-existing environment requirement); this change touches no mobile code, so the host-target lint is not affected.
- **Skipped `cargo clippy --all-features`** — the `mobile` feature on `omnibus-frontend` conflicts with the default `web` feature (pre-existing on main; `use_playback` / `PlaybackState` gated `#[cfg(not(feature = "mobile"))]`). `just lint`'s explicit per-feature invocations (which I ran above) are the intended matrix.
- Sanitizer scope: the change permits `<button>` as a tag on the ammonia allowlist, but only with the spoiler-shaped attribute set — any hand-authored `<button class="…" type="submit" onclick="…">` degrades to a bare `<button>` (still no interactive JS, since the toggle handler only matches `.spoiler`). Locking down bare `<button>` via a `drop_non_spoiler_buttons` guard (analogous to `drop_non_checkbox_inputs`) was deliberately left for a follow-up so this PR stays scoped to the a11y gap.

Closes #644

---
_Generated by [Claude Code](https://claude.ai/code/session_01YMQ1M4kQQA3SmTkRCXFv61)_


seamus-sloan:
Flagging both Copilot warnings for your call before merging — I'm holding off on pushing.

**Warning 1 (line 40 — bare `<button>` slips the allowlist).** Correct. `ammonia` strips disallowed attrs but keeps allowed tags, so a hand-authored `<button class="spoiler">…</button>` without `type="button"` survives sanitization. The PR body deliberately deferred a `drop_non_spoiler_buttons` guard (analogous to `drop_non_checkbox_inputs`) as a follow-up — so this is a known gap, not a surprise, but the follow-up should land before the reveal handler ships in prod. Small guard, low risk to add here if you want it in-scope.

**Warning 2 (line 131 — nested interactive HTML).** Also correct, and it's a real correctness bug, not a nit. `wrap_spoilers` emits `<button>…</button>` before pulldown-cmark renders the inner markdown, so `||[link](url)||` produces `<button><a>link</a></button>`. That's invalid per HTML5 (an interactive element cannot contain another), and browsers/screen readers behave inconsistently — some announce twice, some drop the inner control, some make the link unclickable.

The fix is a design reversal: go back to `<span class="spoiler" role="button" tabindex="0" aria-expanded="false">` with an explicit `onkeydown` for Enter/Space (the original plan alternative). That keeps a11y intact and avoids the nested-interactive HTML.

Not pushing unilaterally since the button-vs-span choice is the whole shape of the PR — want your call on whether to reverse for this PR or land as-is and follow up. I can wire the span+keydown version in a follow-up commit once you say the word.

---

PR #685, opened 2026-07-02
refactor: batch audiobook sync New/Removed buckets to eliminate N+1 queries

seamus-sloan:
## Summary

Closes #632.

Mirror the Changed-bucket batching pattern into the New and Removed buckets of the audiobook sync path (`db/src/sync/audiobooks.rs`). Both buckets used to fire 1–2 DB queries per book while `sync_audiobooks_changed` was already batched.

### `sync_audiobooks_new`
Pre-fetches every candidate `scan_key` in one chunked `SELECT scan_key, id, uuid FROM books WHERE library_id = ? AND scan_key IN (…)` (499 keys per chunk under SQLite's 999-bind cap) into a `HashMap` before the per-book loop, replacing the per-book `SELECT id, uuid FROM books WHERE library_id = ? AND scan_key = ?` `fetch_optional`. The rewrite-in-place branch (fileless-returning + `replace_books` re-add) now looks up in memory; the cross-format attach heuristic path is unchanged.

### `sync_audiobooks_removed`
Introduces `batch_mark_audiobooks_files_missing` — one chunked `DELETE FROM book_files WHERE book_id IN (…)` plus one chunked `UPDATE books SET is_missing_files = 1, missing_files_since = unixepoch() WHERE id IN (…) AND is_missing_files = 0 AND is_missing_files_override = 0` per 500 ids, replacing the per-book `mark_book_files_missing(tx, id)` loop. Same fileless-book semantics (F2), same idempotency guards, one round-trip per chunk instead of two per book.

## Tests

- Extended `sync_audiobooks_with_removed_above_bind_cap_succeeds` (the existing 1000-audiobook wholesale-removal test) to assert every affected row is flagged `is_missing_files = 1` after the batched Removed pass — regression guard on the new `DELETE ... IN (…)` / `UPDATE ... IN (…)` pair.
- Added `sync_audiobooks_new_relinks_returning_group_to_same_uuid_after_fileless_gap`, the audiobook counterpart of the existing ebook F2 relink test, guarding the batched New scan_key lookup against regressing the fileless-then-returning cycle (returning group must relink to the same `books.uuid`, not mint a fresh one).

## Test plan

- [ ] `just test` (CI) — new tests pass; existing audiobook sync tests continue to pass.
- [ ] `just lint` (CI) — `cargo fmt --check` + clippy across the matrix.

## Verification note

Local `cargo test -p omnibus-db` could not be run in this sandbox: the git-sourced `dioxus` dep (`https://github.com/DioxusLabs/dioxus?tag=v0.7.9`) is denied by the egress proxy (403), which blocks any workspace-wide cargo resolution because the workspace `[patch.crates-io]` applies even to the DB crate. `cargo fmt -p omnibus-db --check` is clean. CI is the source of truth here.

---
_Generated by [Claude Code](https://claude.ai/code/session_015XGkb1sHd7fp3UR5VC9Vkw)_

---
### Adversarial review

**Verdict: approved**

Scope is precisely the two files the issue calls out (`db/src/sync/audiobooks.rs`, `db/src/sync/tests.rs`); no scope creep, no migrations touched, no unrelated churn.

- **Acceptance A (New bucket batched).** `sync_audiobooks_new` now pre-fetches `(id, uuid)` for every candidate `scan_key` into a `HashMap` via one chunked `SELECT scan_key, id, uuid FROM books WHERE library_id = ? AND scan_key IN (…)` before the per-book loop; the previous per-book `fetch_optional` is gone. The rewrite-in-place branch (fileless-returning + `replace_books` re-add — preserves `books.uuid`) runs *before* the cross-format attach heuristic, exactly as before. Chunk cap 499 + 1 library_id bind = 500 binds, well under 999.
- **Acceptance B (Removed bucket batched).** `batch_mark_audiobooks_files_missing` replaces the per-book `mark_book_files_missing(tx, id)` loop with one chunked `DELETE FROM book_files WHERE book_id IN (…)` + one chunked `UPDATE books SET is_missing_files = 1, missing_files_since = unixepoch() WHERE id IN (…) AND is_missing_files = 0 AND is_missing_files_override = 0` per 500 ids. Both idempotency guards (`is_missing_files = 0`, `is_missing_files_override = 0`) preserved — a re-run keeps the original `missing_files_since` and leaves wishlist rows untouched. Chunk cap 500 binds, under 999.
- **Idempotency / F2 semantics.** Fileless-book invariant intact: `books` row + `merged_uuids` + FTS survive; only `book_files` is dropped. The rewrite-in-place branch in `sync_audiobooks_new` still keys off `(library_id, scan_key)` and reuses the existing `books.id` + `uuid`, so a returning group re-attaches instead of minting a new uuid.
- **Test rigor.** The extended `sync_audiobooks_with_removed_above_bind_cap_succeeds` now asserts `COUNT(*) WHERE is_missing_files = 1 == N` after the batched Removed pass — would fail if the batched `UPDATE ... IN (…)` regressed to a no-op or lost the flag. The new `sync_audiobooks_new_relinks_returning_group_to_same_uuid_after_fileless_gap` walks the full new → remove (fileless, uuid retained) → re-add cycle and asserts the returning group relinks to the *same* `uuid1` — would fail if the batched pre-fetch skipped the rewrite-in-place branch and let `try_attach_new_audiobook` / `insert_new_audiobook` mint a fresh uuid.
- **Rule 05 line cap.** `db/src/sync/audiobooks.rs` is 783 lines (verified against the PR branch tip), under the ~800-line soft cap.
- **Housekeeping.** Conventional-commit title (`refactor:`), branch matches `refactor/632-…`, PR body opens with `Closes #632`. rustfmt + cargo audit + copilot review all green; `clippy + test` and `bundle` still running at review time — CI is the source of truth per the PR body's verification note (git-sourced `dioxus` dep is proxy-denied locally).

No changes requested.

### DIFF
--- a/db/src/sync/audiobooks.rs
+++ b/db/src/sync/audiobooks.rs
@@ -107,8 +107,8 @@ pub async fn sync_audiobooks_with_progress(
     Ok(())
 }
 
-/// Apply the Removed bucket (F2): resolve affected ids and, via
-/// `books::mark_book_files_missing`, drop the `book_files` rows (parts/chapters
+/// Apply the Removed bucket (F2): resolve affected ids and, in a pair of
+/// chunked batch statements, drop the `book_files` rows (parts/chapters
 /// cascade) and flag each book missing — but **retain** the `books` row, its
 /// links, FTS, and soft-ref user data, so the book stays in browse/search (the
 /// grid hides it via `EXISTS book_files`) and a returning group re-attaches via
@@ -140,9 +140,7 @@ async fn sync_audiobooks_removed(
         }
         let ids = q.fetch_all(&mut **tx).await?;
         missing += ids.len();
-        for id in ids {
-            super::books::mark_book_files_missing(tx, id).await?;
-        }
+        batch_mark_audiobooks_files_missing(tx, &ids).await?;
     }
     if missing > 0 {
         tracing::info!(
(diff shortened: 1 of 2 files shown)


seamus-sloan:
## CI: `clippy + test` red — diagnosis + fix

Compile error, single line:

```
error: unused import: `mark_book_files_missing`
  --> db/src/sync/books/mod.rs:21:31
```

This PR removes the last user of `super::books::mark_book_files_missing` (the audiobook Removed bucket now batches inline) but the `pub(super) use shared::{clear_missing_files_flag, mark_book_files_missing, materialize_new_covers};` re-export in `db/src/sync/books/mod.rs:20-22` still lists it. `sync/books/removed.rs` calls `super::shared::mark_book_files_missing` directly, so the re-export is dead.

**Fix (one line):**

```rust
pub(super) use shared::{clear_missing_files_flag, materialize_new_covers};
```

I prepared this fix locally and verified no other caller uses the removed re-export, but I can't push it from this session — org egress policy allows pushes only to the designated dev branch, not to feature branches. Please apply the change above; CI should go green.

Duplicate-PR note: PR #689 (`fix/632-batch-audiobook-sync-new-removed`) covers the same underlying #632 issue and is currently green. If you're planning to keep one and close the other, this fix is unnecessary — closing this PR resolves it.

---

PR #788, opened 2026-07-06
fix: stop leaking raw DB/internal error text on the RPC surface (#761)

seamus-sloan:
## Summary
- The Dioxus server-fn (`/api/rpc/*`) surface stringified raw `sqlx`/internal error `Display` text straight into `ServerFnError`, so the browser could see SQL/schema/transport details on core reading/annotation flows (journals, highlights, ratings, progress) plus account/authors/settings/kindle. `server/src/backend.rs` already has this discipline on the REST/mobile surface via a shared `internal()` helper; the RPC surface had no equivalent.
- Added `rpc::internal_error()` (mirrors `server::backend::internal`) that logs the real error server-side via `tracing::error!` and returns a fixed `"internal server error"` message to the client. Applied it to every cited leak site in the issue, plus the equivalent `?`-propagated leaks in the same modules (same error types, same functions) that the issue's own evidence list missed because it was grepped for explicit `.to_string()` calls rather than bare `?`.
- `frontend/src/rpc/bookmarks.rs` had already independently fixed this exact pattern with a private, file-local helper — hoisted it into the shared `rpc::internal_error` (used by 9 modules now) instead of leaving 9 near-duplicate copies.
- Left `frontend/src/rpc/books.rs:84` (`PageCursor::decode`) unchanged: on inspection, `CursorError::Malformed`'s `Display` is a fixed, safe `"malformed pagination cursor"` string carrying no internal detail — it already satisfies the acceptance criteria's own carve-out ("errors that are legitimately meant for the user... continue to surface their specific message").

## Test plan
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -p omnibus-frontend --features server --all-targets -- -D warnings` clean
- [x] `cargo clippy --workspace --exclude omnibus-mobile -- -D warnings` clean
- [x] `cargo test -p omnibus-frontend --features server` — 202 passed
- [x] `cargo test -p omnibus` — 297 passed
- [x] `cargo test -p omnibus-db` — passed
- [x] `cargo test -p omnibus-shared` — 92 passed

## Version
- [x] **Patch** — default, unlabeled

## Notes
- Out of scope, flagged for a follow-up issue: `frontend/src/rpc/books.rs`, `authors.rs` (aside from the one function touched here), and `settings.rs` have additional `?`-based calls (e.g. `db::get_settings(...).await?`, `db::list_authors(...).await?`) that funnel through the same blanket `CapturedError: From<E: Into<anyhow::Error>>` conversion and could leak in the same way. They weren't named in #761's evidence or severity discussion (which calls out journals/highlights/ratings/progress specifically), so left untouched here to keep this PR bounded to the cited scope.

---
_Generated by [Claude Code](https://claude.ai/code/session_018P4LFLXa7sMVeVc1CxmXbR)_

### DIFF
--- a/frontend/src/rpc/account.rs
+++ b/frontend/src/rpc/account.rs
@@ -8,7 +8,7 @@ use dioxus::prelude::*;
 use omnibus_db::{self as db, auth::AuthError};
 
 #[cfg(feature = "server")]
-use super::{AuthUser, PoolExt};
+use super::{internal_error, AuthUser, PoolExt};
 
 /// Set (or clear, with `None`/blank) the authenticated user's Kindle email.
 /// Rejects a malformed address with a validation `ServerFnError`.
@@ -17,6 +17,6 @@ pub async fn rpc_set_kindle_email(email: Option<String>) -> Result<()> {
     match db::auth::set_kindle_email(&pool.0, user.id, email.as_deref()).await {
         Ok(()) => Ok(()),
         Err(AuthError::Validation(msg)) => Err(ServerFnError::new(msg).into()),
-        Err(e) => Err(ServerFnError::new(e.to_string()).into()),
+        Err(e) => Err(internal_error("set_kindle_email", e).into()),
     }
 }
--- a/frontend/src/rpc/authors.rs
+++ b/frontend/src/rpc/authors.rs
@@ -20,7 +20,7 @@ use omnibus_shared::AUTHOR_PHOTO_URL_MAX_LEN;
 use omnibus_shared::detect_image_format;
 
 #[cfg(feature = "server")]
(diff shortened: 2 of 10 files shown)


seamus-sloan:
Closing as a duplicate: #789 fixed the same issue (#761) independently and was merged into `main` first (at 2026-07-06T10:35:51Z). Both PRs took the same approach (a shared `internal_error`-style helper mirroring `server::backend::internal`, applied across the same RPC modules, including the same `books.rs:84` carve-out reasoning for `CursorError::Malformed`), so there's nothing here worth cherry-picking on top of what already landed. No further action needed on this branch.

---

PR #792, opened 2026-07-06
refactor: extract shared multipart image-validation pipeline (#766)

seamus-sloan:
## Summary
- Two REST handlers (`post_ebook_cover` in `server/src/backend/overrides.rs` and `put_author_photo` in `server/src/backend/author_photos.rs`) each hand-rolled the same staged multipart image-validation pipeline: content-type check → SVG reject → 10 MiB size cap → magic-byte sniff.
- Extracted the pipeline into a new `extract_validated_image(multipart, field_name)` helper in `server/src/backend/image_upload.rs`, called from both handlers. Behavior is unchanged — same status codes, same user-facing messages (parameterized by field name: `"cover"` vs `"photo"`), same 10 MiB cap, same SVG rejection, same magic-byte sniff via `detect_image_format`.
- Internal-only tracing context strings for the two `field.bytes()`-read-error paths were consolidated from `"read cover field"`/`"read photo field"` to a single `"read image field"` — this only affects the server log line on a 500, not any externally observable response.

## Test plan
- [x] Added `server/src/backend/image_upload/tests.rs` — unit tests for the new helper covering: valid PNG accepted, missing field (400), non-image content-type (400), SVG rejected (400), oversized upload (400), unrecognized magic bytes (415).
- [x] Existing `server/src/backend/overrides/tests.rs` and `server/src/backend/author_photos/tests.rs` pass unchanged.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy -p omnibus --all-targets -- -D warnings` clean.
- [x] `cargo test -p omnibus` — full suite (303 tests) passes.

## Version
- [x] **Patch** — the default; internal refactor, no behavior change, no release-note-worthy feature.

## Notes
- Closes #766.
- No migrations, no API/behavior changes, no Playwright-relevant markup touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FejFgWsFa7Kcxo9ADRnxvS)_

### DIFF
--- a/server/src/backend.rs
+++ b/server/src/backend.rs
@@ -28,6 +28,7 @@ mod covers;
 mod ebooks;
 mod health;
 mod highlights;
+mod image_upload;
 mod journals;
 mod kindle;
 mod overrides;
--- a/server/src/backend/author_photos.rs
+++ b/server/src/backend/author_photos.rs
@@ -16,6 +16,7 @@ use omnibus_shared::detect_image_format;
 use serde::Deserialize;
 use sqlx::SqlitePool;
 
+use super::image_upload::extract_validated_image;
 use super::{internal, AppState};
 use crate::auth::{AdminUser, MediaAuthUser};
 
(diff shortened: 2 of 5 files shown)


seamus-sloan:
Re: the 4 "CRITICAL" Copilot comments on `image_upload/tests.rs` (lines 73, 84, 96, 108) claiming `assert_eq!(body, "...")` won't compile because `Bytes` can't be compared to `&str`:

This is a false positive. `bytes::Bytes` implements `PartialEq<str>` directly (`bytes-1.11.1/src/bytes.rs:825`) plus a blanket `impl<'a, T: ?Sized> PartialEq<&'a T> for Bytes where Bytes: PartialEq<T>` (line 921), which together give `Bytes: PartialEq<&str>` — exactly what `assert_eq!` needs. The pattern already exists unchanged elsewhere in this codebase at `server/src/backend/settings/tests.rs:227` (`assert_eq!(body, "internal server error")`), which is in fact the exact file/line Copilot cited as the reference to follow — it just misread its own citation.

Locally verified: `cargo clippy -p omnibus --all-targets -- -D warnings` and `cargo test -p omnibus` (303 tests, including these four assertions) both pass clean. No change needed.

---

PR #793, opened 2026-07-06
refactor: dedupe multipart image-validation pipeline (#766)

seamus-sloan:
## Summary
- `post_ebook_cover` (`server/src/backend/overrides.rs`) and `put_author_photo` (`server/src/backend/author_photos.rs`) each hand-rolled the same staged multipart image-validation pipeline (content-type check → SVG reject → 10 MiB size cap → magic-byte sniff via `detect_image_format`).
- Factored the shared logic into `extract_validated_image` in a new `server/src/backend/image_upload.rs` module. Each handler now just calls `extract_validated_image(&mut multipart, "cover" | "photo")` and returns the error `Response` on failure.
- Error message text is unchanged (`"{field} must be an image"`, `"SVG {field}s are not accepted"`, `"{field} must be under 10 MB"`, `"missing '{field}' field in multipart body"`, `"Could not detect image format"`) — pure extraction, no behavior change. The one internal log-context string that changed (`"read cover field"`/`"read photo field"` → `"read multipart image field"`) is server-side tracing only, never sent to the client.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy -p omnibus --all-targets -- -D warnings`
- [x] `cargo clippy` (workspace default-members)
- [x] `cargo test -p omnibus` — 297 passed, including all existing `backend::overrides` and `backend::author_photos` cover/photo-upload tests unchanged
- [x] Independent adversarial review pass (separate agent) diffed the extracted helper against the removed inline loops line-by-line, confirmed no scope creep, no migrations touched, and re-ran the full check suite — no findings.

## Version

## Notes
- Pure refactor, no schema/migration changes, no API contract changes.
- Pre-existing gap (not introduced here, out of scope for this issue): neither `overrides/tests.rs` nor `author_photos/tests.rs` has an explicit SVG-rejection test case. Worth a follow-up if desired.

Closes #766


seamus-sloan:
Closing — superseded by #792, which merged the same fix (extracting `extract_validated_image` into `server/src/backend/image_upload.rs`) to `main` about 20 minutes before this PR was opened. Issue #766 is already closed via #792. Confirmed via `git merge-tree` that this branch genuinely conflicts with `main` on `image_upload.rs`/`overrides.rs`/`author_photos.rs` — not worth resolving since it's duplicate work.

---

PR #898, opened 2026-07-10
ci: run omnibus-frontend --features mobile tests (#867)

seamus-sloan:
## Summary
- Adds a `cargo test -p omnibus-frontend --features mobile` step to the `check` job in `.github/workflows/rust.yml`, alongside the existing `server`/`shared`/`db` test steps.
- The `mobile` feature exercises a distinct cfg-split code path from `server` (`frontend/src/data.rs`'s `token_store`/`app_dirs` native token-persistence, plus the `mobile`-gated `DataError` variants) that had zero CI signal — even though `Justfile`'s `test` recipe has run it locally all along (`just test` → `cargo test -p omnibus-frontend --features mobile`).
- The `mobile` feature only pulls in `reqwest`/`tokio`/`jni`/`ndk-context` (checked `frontend/Cargo.toml`), not the GTK/wry desktop backend that `omnibus-mobile` needs — so the new step runs in the plain `nix develop` shell like the other test steps, not `.#mobile`. Confirmed locally (189 tests pass under this feature set).
- One-line doc fix in `CLAUDE.md`'s quick-commands section: `just test`'s comment said `frontend(--features server)` but the recipe has run both `server` and `mobile` feature sets for a while — updated to match reality.
- Closes #867

## Test plan
- [x] `cargo test -p omnibus-frontend --features mobile` — 189 passed, 0 failed (ran directly; this sandbox has no Nix, so `nix develop --command cargo test ...` itself wasn't invoked, but the underlying command is identical).
- [x] `cargo fmt --check` — no changes needed (no `.rs` files touched).
- [x] YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`).
- [ ] CI: this PR's own `check` job run is the first real exercise of the new step end-to-end (Nix shell, cache, etc.) — watching for green.

## Version
- [ ] **Minor** — add the `minor version` label (`vX.Y.Z` → `vX.(Y+1).0`).
- [ ] **Patch** — the default; add the `patch version` label or leave unlabeled
  (`vX.Y.Z` → `vX.Y.(Z+1)`).
- [x] **No release** — CI-only + docs change; per the template, unlabeled PRs touching only `.github/` and `*.md` skip the release automatically, so left unlabeled.

## Notes
- No Rust code changed — CI workflow + one doc line only.
[...]

PR #900, opened 2026-07-10
test: cover closed-pool DB-passthrough paths for ~20 db thiserror enums (#856)

seamus-sloan:
## Summary
- Adds a `..._propagates_db_error_when_pool_is_closed`-style test (the existing pattern in `identity/tests.rs`, `journals/tests.rs`, `missing_files/tests.rs`, `normalize/tests.rs`, `ratings/tests.rs`) for each untested `Db`/`Sqlx`-passthrough `thiserror` variant cited in #856: `covers`, `browse`, `author_photos_data`, `bookmarks`, `discovery`, `highlights`, `hls`, `indexer`, `palette`, `progress`, `settings`, `shelves`, `sort_keys`, `suggestions::data`, `kepub`, `kindle`, and `sync::books::SyncError` (via a dropped-table transaction, since that enum has no direct pool-level entry point — it's produced deep inside the audiobook chapter writer).
- `suggestions::hardcover::HardcoverError::Http` isn't a DB variant but was grouped in the same issue entry — covered with an unreachable-endpoint test (`http://127.0.0.1:1`) using the existing `wiremock` harness pattern already in `suggestions/tests.rs`.
- Test-only change — no production code touched.
- Closes #856

## Three cited variants are left untested, with a note instead of a fabricated test
- **`ThumbError::Db`** (`db/src/thumbs.rs`) — dead code. No function in the module ever takes a `SqlitePool`, so this variant is unreachable from any call site today.
- **`AudiobookError::Io`** (`db/src/audiobook/parse.rs`) — already covered by the pre-existing `audiobook_error_io_variant_renders_useful_message` test (verifies the `#[from]` conversion + `Display` message). No real IO-failure call site exists in the public API (`lofty::read_from_path` failures map to `Tag`, not `Io`) to hang a closed-pool-style test off of.
- **`InitDbError::{Normalize,Identity,SortKeys,MissingFiles}`** (`db/src/pool.rs`) — these boot-sequence wrapping arms only fire when an individual backfill fails *after* migrations already succeeded on the same connection. Reaching that isn't possible without adding a test seam to `init_db` itself, which is out of scope for a test-only PR.
- **`frontend::DataError::{Network,Decode}`** — compile only under `feature = "mobile"`/`"web"`, but `frontend/Cargo.toml`'s `mobile` feature enables only tokio's `sync`/`time` subfeatures, not `macros`/`rt-multi-thread` (only `server` does) — so no async test runtime is available to drive a `reqwest` call under `--features mobile` alone. This is adjacent to #867 (already being addressed in open PR #898, which wires `--features mobile` into CI) but fixing it here would mean changing `Cargo.toml` feature gates, which is out of scope for this issue.

## Test plan
- [x] `cargo test -p omnibus-db` — 895 passed, 1 pre-existing failure unrelated to this change (see Notes)
- [x] `cargo test -p omnibus-frontend --features server` — 226 passed, 0 failed
- [x] `cargo test -p omnibus-shared` — 99 passed, 0 failed
- [x] `cargo clippy -p omnibus-db --all-targets -- -D warnings` — clean
- [x] `cargo clippy` (default-members) — clean
- [x] `cargo fmt --check` — clean

## Version
  (`vX.Y.Z` → `vX.Y.(Z+1)`).
- [ ] **No release** — add the `no release` label to skip cutting a release.

## Notes
- Pre-existing, unrelated failure on `main`: `omnibus::backend::uploads::tests::commit_rejects_read_only_library_with_400` fails in this sandbox because it runs as `root`, which bypasses the `chmod 0o555` read-only check the test relies on to force a `PermissionDenied`. Not touched by this PR (server crate, no files in this diff) — flagging per repo convention rather than silently working around it.
- No migrations, schema, or RPC-surface changes. `Cargo.lock` untouched.

### DIFF
--- a/db/src/author_photos_data/tests.rs
+++ b/db/src/author_photos_data/tests.rs
@@ -293,3 +293,11 @@ async fn delete_author_rebuilds_fts_for_all_affected_books_in_batch() {
         "deleted author must not appear in any FTS row"
     );
 }
+
+#[tokio::test]
+async fn get_author_photo_propagates_db_error_when_pool_is_closed() {
+    let pool = init_db("sqlite::memory:").await.unwrap();
+    pool.close().await;
+    let err = get_author_photo(&pool, 1).await.unwrap_err();
+    assert!(matches!(err, AuthorPhotosDataError::Db(_)));
+}
--- a/db/src/bookmarks/tests.rs
+++ b/db/src/bookmarks/tests.rs
@@ -263,3 +263,21 @@ async fn list_bookmarks_query_plan_uses_covering_index() {
         "expected index-only sort — plan still uses a temp b-tree:\n{plan}",
     );
 }
+
+#[tokio::test]
+async fn create_bookmark_propagates_db_error_when_pool_is_closed() {
+    let pool = init_db("sqlite::memory:").await.unwrap();
+    pool.close().await;
+    let err = create_bookmark(
+        &pool,
+        1,
+        &CreateBookmark {
+            book_uuid: "any-uuid".into(),
+            position: "0".into(),
+            title: None,
+        },
+    )
+    .await
+    .unwrap_err();
+    assert!(matches!(err, BookmarkError::Sqlx(_)));
+}
--- a/db/src/browse/tests.rs
+++ b/db/src/browse/tests.rs
@@ -495,3 +495,11 @@ async fn list_series_includes_series_from_audiobook_library() {
     );
     drop(guard);
 }
+
+#[tokio::test]
+async fn list_authors_propagates_db_error_when_pool_is_closed() {
+    let pool = init_db("sqlite::memory:").await.unwrap();
+    pool.close().await;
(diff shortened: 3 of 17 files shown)


seamus-sloan:
`Mobile E2E / maestro (iOS simulator)` failed on this PR, but it's a CI infra flake, not something introduced by this diff (test-only additions to `db/src/**/tests.rs`, nothing under `mobile/` or `ui_tests/maestro/`). The Rust build, app bundling, and simulator install all succeeded; the failure is Maestro's XCTest driver timing out while attaching to the booted simulator:

```
xcuitest.installer.LocalXCTestInstaller$IOSDriverTimeoutException: iOS driver not ready in time, consider increasing timeout by configuring MAESTRO_DRIVER_STARTUP_TIMEOUT env variable
```

`Rust` and `E2E` both passed on this commit. I don't have permission to trigger a re-run from here (got a 403 re-running the failed job via the API) — could you kick off a re-run of the Mobile E2E workflow when you get a chance?

---

PR #982, opened 2026-07-11
fix: surface journal entry delete errors to the owner (#901)

seamus-sloan:
## Summary
- `BdJournalEntryCard` gated the delete-failure error span on `!is_owner`, but Delete is an owner-only action — so the one user who can ever trigger it never saw the failure (unless already mid-edit, which has its own error UI). Render the error unconditionally in the non-editing branch, matching how the edit form and composer already surface errors regardless of ownership.

## Test plan
- [x] `cargo fmt --check -p omnibus-frontend`
- [x] `cargo test -p omnibus-frontend --features server` (229 passed)
- [x] Added a Playwright regression test (`ui_tests/playwright/tests/flows/journal.spec.ts`) that forces a 500 on `/api/rpc/journals/delete` and asserts the owner's entry card shows the inline alert — this is not runnable in this sandbox (no Nix/Chromium), but type-checks cleanly with `tsc --noEmit` and follows the existing error-path pattern in the same file (see the "failed publish" test directly above it).

## Version
- [x] **Patch** — the default; left unlabeled per the template (bug fix, no behavior addition).

## Notes
- Bug and fix are exactly as scoped in #901 — no scope creep beyond the one conditional and its regression test.

Closes #901

---
_Generated by [Claude Code](https://claude.ai/code/session_015buRcf18BYcVBLpv21c8vP)_

### DIFF
--- a/frontend/src/pages/book_detail/journal/entry_card.rs
+++ b/frontend/src/pages/book_detail/journal/entry_card.rs
@@ -173,10 +173,8 @@ pub(super) fn BdJournalEntryCard(
                     class: "bd-journal-entry-body",
                     dangerous_inner_html: "{entry.body_html}",
                 }
-                if !is_owner {
-                    if let Some(msg) = error() {
-                        span { class: "mono bd-journal-error", role: "alert", "{msg}" }
-                    }
+                if let Some(msg) = error() {
+                    span { class: "mono bd-journal-error", role: "alert", "{msg}" }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


seamus-sloan:
`maestro (iOS simulator)` failure is pre-existing, not caused by this PR — the `connect` flow fails with the same signature already tracked in #902 (`Assertion is false: "http://127.0.0.1:3000" is visible`), and `connect_error` passed. This PR doesn't touch anything in the mobile connect screen or server-URL flow (only `frontend/src/pages/book_detail/journal/entry_card.rs` and a web Playwright spec). Leaving as-is per #902 rather than fixing here — out of scope for this change.

---

PR #991, opened 2026-07-11
fix: resolve current user on mobile so journal edit/delete buttons render (#893)

seamus-sloan:
## Summary
- `use_current_user_summary` (`frontend/src/contexts.rs`) only populated its signal via the web-only `CurrentUser` context. On mobile builds that context doesn't exist, so the signal stayed `None` forever, `is_owner` in `BdJournalEntryCard` was permanently `false`, and the owner-only Edit/Delete action row never rendered — for anyone, on any journal entry, on mobile.
- Added a mobile branch to `use_current_user_summary` that resolves the current user via a direct bearer-authenticated `data::get_me` fetch, mirroring the identical pattern already used by the Account screen's mobile identity fetch (`frontend/src/pages/account.rs`). The rest of the edit/delete flow (`BdJournalEntryEditForm`, the mobile `PATCH /api/journals/{id}` transport, the server handler, the db update) was already correct and platform-agnostic — only the ownership signal was broken.
- Updated a stale comment in `frontend/src/pages/book_detail/journal.rs` that documented the old (buggy) "mobile stays `None`" behavior as intentional.

## Investigation notes
- Issue #893 was initially investigated as a web-only ticket and found to be already fully implemented there (see the first comment on the issue) — the reporter then clarified the gap is mobile-specific, which is what this PR addresses.
- Root cause confirmed by reading `use_current_user_summary`'s web-only `#[cfg(feature = "web")]` effect gate alongside `entry_card.rs`'s `is_owner` derivation; not a markup/hydration cfg-gate issue (rule 07 was already honored — the Edit form's rsx is identical across targets), purely a data-population gap.

## Test plan
- [x] `cargo fmt -p omnibus-frontend --check`
- [x] `cargo clippy -p omnibus-frontend --features mobile --all-targets -- -D warnings`
- [x] `cargo test -p omnibus-frontend --features server` (229 passed)
- [ ] No direct unit test added: `use_current_user_summary` is a Dioxus hook (`use_signal`/`use_effect`/`spawn`) with no existing test harness in this codebase for that class of code — every other hook in `contexts.rs` (`use_is_admin`, etc.) and the analogous mobile `get_me` fetch in `account.rs` are untested the same way; issue #888 ("Add VirtualDom Tests to Augment Maestro Tests") tracks building that harness. Also confirmed there is **no existing Maestro coverage of journals at all** (only `connect`/`connect_error` flows exist under `ui_tests/maestro/flows/`) — adding first-time journal Maestro coverage is a larger, separate effort than this focused fix and is left out of scope here.

## Version
- [x] **Patch** — the default; left unlabeled per the template (bug fix, no behavior addition).

## Notes
- Scope stayed to the exact ownership-signal bug described in #893; no unrelated cleanup.

Closes #893

---
_Generated by [Claude Code](https://claude.ai/code/session_01EG8aGToZGCcNwoyvuuMqTP)_

### DIFF
--- a/frontend/src/contexts.rs
+++ b/frontend/src/contexts.rs
@@ -151,12 +151,13 @@ pub fn use_is_admin() -> Signal<bool> {
     is_admin
 }
 
-/// Derive the resolved current user from the app-wide [`CurrentUser`]
-/// context instead of an independent per-mount `/api/auth/me` fetch,
-/// flattening "not yet resolved" and "unauthenticated" alike to `None`.
-/// Starts `None` so SSR and the first WASM paint agree (rule 07); only the
-/// web build wires the effect that fills it in once the context resolves.
-#[cfg_attr(not(feature = "web"), allow(unused_mut))]
+/// Derive the resolved current user: on web, from the app-wide
+/// [`CurrentUser`] context instead of an independent per-mount
+/// `/api/auth/me` fetch; on mobile (which has no `CurrentUser` context,
+/// only a bearer token), via a direct `data::get_me` call mirroring the
+/// Account screen's mobile identity fetch. Starts `None` so SSR and the
+/// first WASM paint agree (rule 07); the post-mount effect fills it in.
+#[cfg_attr(not(any(feature = "web", feature = "mobile")), allow(unused_mut))]
 pub fn use_current_user_summary() -> Signal<Option<omnibus_shared::UserSummary>> {
     let mut current = use_signal(|| None);
     #[cfg(feature = "web")]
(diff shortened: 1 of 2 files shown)


seamus-sloan:
The `maestro (iOS simulator)` job failure here is the pre-existing flake tracked in #902 — same assertion (`"http://127.0.0.1:3000" is visible`) on the unrelated `connect` flow, nothing to do with this PR's diff (`frontend/src/contexts.rs` / `frontend/src/pages/book_detail/journal.rs`). `connect_error` and the Android job both passed in this run, and `main`'s last three completed Mobile E2E runs (`93809f1`, `b9c3847`, `0f64845`→`02839da`) all passed the iOS job cleanly, confirming this is intermittent rather than a regression from this change.

My GitHub token doesn't have permission to re-run failed jobs (`403` on `rerun-failed-jobs`) — @seamus-sloan, could you re-run the `Mobile E2E` job on this PR when you get a chance? I'll keep watching and report back once it's green.

---

PR #1015, opened 2026-07-12
fix: parse EPUB content documents by manifest media-type, not extension (#1012)

seamus-sloan:
## Summary
- Closes #1012.

`epub.js` (vendored at `frontend/assets/vendor/epub.min.js`) chose its HTML vs XHTML parser from a content document's **file extension**, ignoring the OPF manifest's declared media-type. Tools like Gutenberg's `ebookmaker` name XHTML content documents `.html`, so epub.js HTML-parsed them — mis-parsing self-closing elements (most visibly a self-closing named anchor at each chapter start), which HTML5's adoption-agency reconstruction then leaks into the rest of the chapter body, wrapping it in a href-less anchor.

### Fix (Option A from the issue — patch the vendored parse decision)
- `Spine.unpack()` now threads the OPF manifest's already-parsed media-type (`manifest[idref].type`) onto each spine item and its `Section`.
- `Section.load()` maps that media-type to an explicit parse-type argument (`"xhtml"` for `application/xhtml+xml`, `"html"` for `text/html`) passed to `request()`, instead of letting `request()` infer the type from the URL extension. Any other/undefined media-type still falls through to the prior extension-based inference (no behavior change for non-content-document requests).
- The `xhtml` branch of `request()` now retries as HTML if the XHTML parse yields a `<parsererror>`, so a malformed/mislabeled document degrades to the previous behavior instead of blank-paging (AC2).
- Added a one-line hand-patch comment at the top of the vendored file per convention for patched vendored code, noting it must be re-applied on the next epub.js version bump.

## Test plan
- [x] Real-browser (Chromium via Playwright, `/opt/pw-browsers/chromium`) `DOMParser` check mirroring the exact patched logic against a crafted repro (self-closing named anchor, declared XHTML): confirmed HTML-mode parsing (old behavior) puts the following `<p>` inside the anchor (`oldPInAnchor: true`), while the patched XHTML-mode path does not (`newPInAnchor: false`, `newParseError: false`), and a malformed-markup fallback still yields usable content (`fallbackHasContent: true`) rather than blank-paging.
- [x] `node -e "new Function(fs.readFileSync('epub.min.js'))"` — confirms the patched bundle is still syntactically valid JS.
- [ ] Full manual verification against real Gutenberg (`dracula`, `frankenstein`) and Calibre-sourced (`Empire of the Vampire`, `Oathbringer`) books in the running web reader — **not run in this sandbox** (no network access to fetch those books / no seeded library available in this environment). Recommend a manual spot-check before merge.
- [x] No Rust files changed (`git diff --stat` — vendored JS asset only); `cargo fmt`/`clippy`/`test` unaffected.

## Version
- [ ] **Patch** — default (unlabeled)

## Notes
- Underlying cause behind the link-color mitigation in #1011; the `a[href]` CSS scoping from that PR is unchanged.
- This patches minified vendored code — flagging per repo convention: it must be re-applied (and re-verified) on the next `epub.js` version bump. A short comment at the top of `epub.min.js` documents this.
- I could not reproduce against the real Gutenberg/Calibre books cited in the issue's evidence section since this sandbox has no network access and no seeded library — verification here is limited to a real-Chromium `DOMParser` check of the exact patched decision logic plus a syntax-validity check of the patched bundle. Recommend a manual pass with `ui-validate` against the cited books before merge.


seamus-sloan:
**CI diagnosis — E2E failure (shard 1/2) is pre-existing on `main`, unrelated to this change.**

The only failing test is `tests/flows/journal.spec.ts:409:5 › hides markdown markers on lines away from the caret` (`toHaveCount(2)` on `.cm-line`, received `1`). This PR only touches `frontend/assets/vendor/epub.min.js` (the EPUB reader), which has no relationship to the journal markdown editor.

I confirmed the same test fails with the identical error on `main` at the exact base commit this PR is built from ([run `29176003689`](https://github.com/seamus-sloan/omnibus/actions/runs/29176003689), commit `b0e7e826`) and on the prior `main` commit ([run `29171414668`](https://github.com/seamus-sloan/omnibus/actions/runs/29171414668)) — both before this PR's commit exists. It looks related to the recently-merged `#1008` (journal editor markdown-marker hiding) feature, not this PR.

Per the fixer runbook, I'm leaving this pre-existing failure alone rather than fixing it out-of-scope here — flagging it in case it's worth a separate issue. Everything else on this PR (E2E shard 2, Mobile E2E) is green/in progress.

---

PR #1018, opened 2026-07-12
fix: apply playback rate to mobile audiobook time-left labels (#1000)

seamus-sloan:
## Summary
- On the mobile audiobook player, the "time left" labels (full-player scrubber remaining time, the within-chapter "X left" readout, and the mini-player's "X left" subtitle) were computed as raw audio-content time remaining (`duration - elapsed`), with no adjustment for the current playback rate — so at 1.5x speed the app still reported the same "time left" as at 1x, which reads as wrong once you're actually finishing the book faster.
- Position labels (the elapsed time, the within-chapter position) are correctly left alone — they represent where you are in the book's content, which doesn't change with playback speed.
- Added a small pure helper, `time_left_at_rate(remaining_seconds, rate)`, in `frontend/src/pages/listen/mobile/view.rs` and applied it at the three "time left" display sites in `frontend/src/pages/listen/mobile.rs` (scrubber remaining + chapter countdown) and `frontend/src/pages/listen/mobile/mini.rs` (mini-player subtitle). `rate` was already threaded through `MobilePlayback`/`PlayerProps`, so no new plumbing was needed beyond reading `ctx.rate` in the mini-player.

## Root cause detail (for reviewers)
The issue's screenshot shows `elapsed = 12:37` next to a within-chapter readout of `12:13 / 17:34 · 5:21 left` and a whole-book `-11:53:03`, at `1.50×`. The `12:37` vs `12:13` values are *not* a bug — one is whole-book position, the other is within-chapter position, offset by that chapter's ~24s start; they were always different quantities. The actual bug the title points at is that neither `5:21 left` nor `-11:53:03` accounted for the circled `1.50×` rate — both are pure audio-content-time subtractions. This PR leaves position math untouched and only scales the "left"/remaining labels by `1/rate`.

Desktop (`frontend/src/pages/listen/ready_player.rs` / `stage.rs`) has the same unscaled `remaining` computation, but the issue scoped this fix to the two named mobile files, so desktop is left as-is — flagging it here as related follow-up debt if wanted.

## Test plan
- [x] `cargo fmt --check` (repo-wide) — clean
- [x] `cargo clippy -p omnibus-frontend --features mobile --all-targets -- -D warnings` — clean
- [x] `cargo test -p omnibus-frontend --features mobile` — 29 passed, including 2 new `time_left_at_rate` tests (inverse-scaling case + non-positive-rate fallback)
- [x] `cargo clippy -p omnibus-frontend --features server` — clean (unaffected build path, sanity-checked)

## Version

## Notes
- Scope: only the two files cited in #1000 (`mobile.rs`, `mobile/mini.rs`) plus the new pure helper + tests in `mobile/view.rs`. No behavior change to position (elapsed/within-chapter) labels.

Closes #1000

---
_Generated by [Claude Code](https://claude.ai/code/session_0138RaiyyHmLSNyesjEmWvxk)_

### DIFF
--- a/frontend/src/pages/listen/mobile.rs
+++ b/frontend/src/pages/listen/mobile.rs
@@ -31,7 +31,7 @@ mod view;
 use bookmarks_sheet::{use_mobile_bookmarks, BookmarksSheet, MobileBookmarks};
 use sheets::{snap_rate, ChaptersSheet, SleepSheet, SpeedSheet};
 use state::{sleep_pill_label, use_mobile_playback, SleepState};
-use view::{chapter_index_for_elapsed, format_hms, format_ms, PlayerView};
+use view::{chapter_index_for_elapsed, format_hms, format_ms, time_left_at_rate, PlayerView};
 
 pub use host::MobileAudioHost;
 pub use mini::MobileMiniPlayer;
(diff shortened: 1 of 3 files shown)


seamus-sloan:
Closing as a duplicate of #1016, which implements the same #1000 fix with a more robust rate guard (handles NaN/Infinity) and broader test coverage. Superseded — keeping #1016.

PR #1019, opened 2026-07-12
fix: multi-file picker for Read/Listen CTAs (#1005)

seamus-sloan:
## Summary
- Closes #1005
- Adds a small file-picker dropdown (`BdFilePickerMenu`, `frontend/src/pages/book_detail/file_picker.rs`) next to the hero Read/Listen CTAs (`hero.rs`) and the mobile CTA row (`mobile.rs`). It renders only when the book has more than one `book_files` row of the format that CTA opens (a duplicate-format group — e.g. two audiobooks merged via F5.10 that both shipped an MP3, or two EPUB editions), letting the reader pick a specific `file_id` before navigating instead of always landing on the DB's primary-file pick.
- **Scope note / deviation from the issue's premise:** the issue's own example ("an EPUB and a PDF") is actually a *different, already-solved* case — `book_files` is populated only when a book has ≥2 files of the *same* format (`db::books::get::get_book` — `has_multipart`). A single EPUB + single PDF is two different `formats`, already handled by the existing per-format `FormatSwitcher` in the rail. This PR targets the case the data model actually supports: duplicate files of one format.
- **UI shape:** rather than inventing a new picker paradigm, this reuses two existing patterns almost verbatim — the hero `BdExportMenu`'s trigger + scrim + focus-on-mount dialog shape, and the `book_files`/`file_id` query-param plumbing the rail's `FormatSwitcher::MultiFileRow` already established (server-side resolution via `?file_id=N` was already wired for both `/api/ebooks/:uuid/file` and `/api/audiobooks/:uuid/manifest`; nothing changed there).
- `frontend/src/pages/book_detail/mod.rs` gains the new (feature-ungated) `file_picker` submodule alongside `journal`/`rating`, since the picker is shared between the web hero and the mobile CTA row, not web-only.
- `frontend/assets/atrium.css` gets a small block of new classes that lean on the existing `.bd-export-*` scrim/panel/item look.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings` (default members) and `cargo clippy -p omnibus-frontend --features {server,web,mobile} --all-targets -- -D warnings`
- [x] `cargo test -p omnibus-frontend --features server` — 233 passed, including 4 new `file_picker` unit tests
- [x] `cargo test -p omnibus-db` — 921 passed
- [x] `cargo test -p omnibus` — 319 passed, 1 failed (`commit_rejects_read_only_library_with_400`) — **pre-existing, unrelated to this change**: it chmods a temp dir `0o555` and expects a `PermissionDenied`, which doesn't fire when the test runs as root (root ignores the write-permission bit). Untouched by this PR (no changes under `server/src/backend/uploads/`).
- [x] `npx stylelint --config .stylelintrc.json 'frontend/assets/*.css'` — clean
- [x] Added Playwright coverage for AC3 (`ui_tests/playwright/tests/flows/book_detail.spec.ts`): merges two same-format MP3 audiobook fixtures via the merge RPC directly (the merge dialog UI itself is already covered by `merge.spec.ts`) to produce a real duplicate-format `book_files` group, opens the new Listen picker, confirms two entries, and asserts picking the non-default file navigates to `/listen/:uuid?file_id=N` with the manifest fetch carrying that id. Also asserts AC2 (no picker for a single-file audiobook) inline in the existing "Start listening" test.
- [ ] **Could not execute Playwright or a live browser check in this sandbox** — no `nix`/`dx` (dioxus-cli) available to build the WASM client bundle. Confirmed hands-on: a plain `cargo run -p omnibus` panics on a missing `target/debug/public` (the `dx`-produced asset dir), and even after stubbing that directory in, the book-detail page's content is entirely client-fetched (SSR renders an empty shell), so there was nothing to inspect without a working WASM bundle either way. `npx playwright test --list` does parse and list the new spec correctly. **Needs a human (or an environment with the `web`/`e2e` Nix shells) to run `ui-validate` and `npx playwright test` for real** before merge.

## Version
- [ ] **Minor**
- [x] **Patch** — leaving unlabeled (default)
- [ ] **No release**

## Notes
- AC1 (picker appears for >1 file), AC2 (no picker for single-file books), and AC3 (Playwright coverage) are all implemented; AC1/AC2 are verified by code review + the hydration-parity check (no `cfg`-gated rsx bodies — only the post-mount focus effect is gated, matching `export_menu`/`user_menu`'s existing pattern) but not by an actual browser render, per the sandbox limitation above.
- Mobile (`mobile.rs`) got the same picker wired into its CTA row for consistency, but mobile builds under a different feature and isn't reachable by Playwright (WKWebView isn't CDP-inspectable) — verified only via `cargo clippy -p omnibus-frontend --features mobile --all-targets`, not a real device/simulator render.


seamus-sloan:
The `playwright (shard 1/2)` failure here (`journal.spec.ts:409 "hides markdown markers on lines away from the caret"`) is a pre-existing flake, not caused by this PR — it fails identically on `main` at the same base commit (`b0e7e826`, [run 29176003689](https://github.com/seamus-sloan/omnibus/actions/runs/29176003689/job/86605893869)), unrelated to the `book_detail` hero/mobile changes here. No action needed on this PR for it.

---

PR #1211, opened 2026-07-21
chore: extract Playwright report publishing into a composite action

seamus-sloan:
## Summary
- Extract the `e2e.yml` `publish-report` job's steps (sticky PR comment, blob-report merge, gh-pages publish) into a reusable composite action at `.github/actions/publish-playwright-report`.
- The job keeps its `permissions`, `gh-pages-publish` concurrency, and run-on-failure `if`; it now just checks out and calls the action with `github-token` + `pages-base-url`.
- Everything repo-specific is a parameterized input whose default matches today's value, so the action can be reused from another repo (`playwright-dir`, `gh-pages-branch`, `main-branch`, `report-subdir`, `report-title`, `blob-report-pattern`, `node-version`).
- Add the new action to the `changes` path filter so a PR editing only it still gets E2E coverage.

## Test plan
- [x] Both `action.yml` and `e2e.yml` parse as valid YAML locally.
- [ ] E2E workflow (triggered by the `e2e.yml` change) goes green; `publish report` posts the sticky report-URL comment and the linked report resolves.

## Version
By default a merge to `main` cuts a patch release. Pick the version update this
PR should trigger; labels override the default:
- [ ] **Patch** — the default; add the `patch version` label or leave unlabeled
  (`vX.Y.Z` → `vX.Y.(Z+1)`).
- [x] **No release** — add the `no release` label to skip cutting a release.

If both `minor version` and `patch version` are applied, `minor version` wins.
Unlabeled PRs that only touch docs (`*.md`, `docs/`) or CI (`.github/`) skip the
release automatically.

## Notes
- CI-only change; omnibus behavior is preserved. One cosmetic generalization: the meta-refresh redirect and the (dead) "no main report yet" landing pages now use the generic `report-title` input ("Playwright report") in their `<title>`/heading instead of the hardcoded "Omnibus — …". The PR-comment and run-summary headings are byte-identical to before.


.github/actions/publish-playwright-report/action.yml
```diff
@@ -0,0 +1,122 @@
...
+      with:
+        path: all-blob-reports
+        pattern: ${{ inputs.blob-report-pattern }}
+        merge-multiple: true
+
+    - name: Merge blob reports into one HTML report
+      shell: bash
+      env:
+        PLAYWRIGHT_DIR: ${{ inputs.playwright-dir }}
+      run: bash "$GITHUB_ACTION_PATH/merge-report.sh"
```
seamus-sloan: It should instead just be a requirement for the `publish-report` action to receive a merged report file instead.

Outside of this action, we should merge the report and then pass the merged report into the `publish-report` action.

.github/actions/publish-playwright-report/merge-report.sh
seamus-sloan, line 1: Delete this.

.github/actions/publish-playwright-report/publish-report.sh
```diff
@@ -0,0 +1,76 @@
...
+if git clone --depth 1 --branch "$GH_PAGES_BRANCH" "$AUTH_URL" gh-pages 2>/dev/null; then
+  cd gh-pages
+else
+  git clone --depth 1 "$AUTH_URL" gh-pages
+  cd gh-pages
+  git checkout --orphan "$GH_PAGES_BRANCH"
+  git rm -rfq . || true
+fi
+git config user.name "github-actions[bot]"
+git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
seamus-sloan: Why do these need to be specifically these values? Can we at least set these strings as variables at the top of the file?

.github/actions/publish-playwright-report/action.yml
seamus-sloan: Done in e7dcb8cd — the action now takes a required `report-dir` input and only comments + publishes. Download + `merge-reports` moved to the e2e.yml `publish-report` job, which produces the merged report and passes it in, so the action no longer needs Node/Playwright.

.github/actions/publish-playwright-report/merge-report.sh
seamus-sloan: Deleted in e7dcb8cd — merging now happens in the workflow, not the action.

.github/actions/publish-playwright-report/publish-report.sh
seamus-sloan: Fixed in e7dcb8cd — hoisted to `COMMIT_USER_NAME` / `COMMIT_USER_EMAIL` at the top of the file. Those are the canonical GitHub Actions bot identity (name + `41898282+github-actions[bot]@users.noreply.github.com`), so the auto-generated report commits are attributed to the bot rather than a person.

PR #1235, opened 2026-07-23
feat: Apple Books-style reader chrome for the mobile ereader

roberte777:
## What

Reworks the ereader chrome — primarily for the mobile app, with the web reader kept consistent — to match the Apple Books reading experience.

### Seamless page ground (no more black bezels)
The reader surface is now painted with the per-theme epub page color (a new `--rd-page` token mirroring the glue's registered body backgrounds: `#201e1b` dark / `#fcfbfa` light / `#ede4d0` sepia), so the strips where the bars live are indistinguishable from the page once the chrome hides. The bars themselves keep their `--bg-0` tint, so summoned menus still read as chrome. The loading/error overlay follows the page color too.

### Slim ambient footer + minimal-chrome labels (phone breakpoint)
- The bottom band slims from 54px to 30px via a shared `--rd-bottom-h` variable — the footer, the stage inset, the immersive (docked audiobook) inset, and the mini-player's pinned position all follow it, so the bottom stack stays aligned.
- Hiding the chrome swaps content instead of leaving empty strips: a small centred book-title label fades in where the toolbar was, and the footer shows only the bare centred page number (new `format_ambient_page`). The richer readout (`p. 4 of 845 · 0%` / `Ch 4 of 32`) rides with the summoned menus.
- The native app now opens books with the chrome hidden (`cfg!(feature = "mobile")` seed — SSR/web still seed visible, keeping hydration parity per rule 07).

### No stranded readers on web
The hide-transforms are now gated to `@media (max-width: 480px), (hover: none) and (pointer: coarse)` (the same gate as the page-turn gutters). A mouse-only viewport always shows the bars — hiding the chrome at phone width and resizing up can no longer leave the reader with no way to summon the menus. Touch tablets (iPad) keep tap-to-hide since a tap there can always bring the chrome back.

### Instant page numbers on re-open
epub.js recomputed its whole-book "locations" pagination on every open — seconds on desktop, much slower in the mobile WebView, during which no page number exists. The glue now caches the generated locations per book in localStorage (keyed by the new `locationsKey` init opt, the book uuid), threaded through both the web bootstrap and the mobile init opts. First open still generates once; every open after shows the page number as soon as the book renders (~0.8s measured, full page load included).

## Testing

- `just lint` (fmt + clippy matrix incl. mobile + stylelint) clean
- `cargo test -p omnibus-frontend --features server`: 370 passed (adds `format_ambient_page` + bootstrap `locationsKey` coverage)
- Playwright `reader.spec.ts`: 10/10, incl. the phone-viewport tests
- Browser-verified across dark/sepia/light at desktop + phone widths: measured the ambient title/page centring (dead-centre), the 30px footer geometry (no prose overlap), the desktop gate (`rd-chrome-hidden` inert on a mouse viewport), and the locations-cache reload timing

### DIFF
--- a/frontend/assets/atrium.css
+++ b/frontend/assets/atrium.css
@@ -5273,8 +5273,13 @@ h1 { font-size: 1.4rem; margin-bottom: 0.5rem; }
   padding: env(safe-area-inset-top) 32px 0; align-items: center;
   justify-content: center; pointer-events: none; overflow: hidden;
   white-space: nowrap; font-size: 13px; color: var(--ink-3);
-  opacity: 0; transition: opacity .22s ease; }
-.rd-chrome-hidden .rd-ambient-title { opacity: 1; }
+  /* `visibility` rides the fade so the inactive label leaves the
+     accessibility tree (opacity alone keeps it exposed and screen readers
+     would announce the title twice); the markup is `aria-hidden` too since
+     the label only mirrors the toolbar title visually. */
+  opacity: 0; visibility: hidden;
+  transition: opacity .22s ease, visibility .22s; }
+.rd-chrome-hidden .rd-ambient-title { opacity: 1; visibility: visible; }
 /* Phone minimal-chrome footer counterpart to `.rd-ambient-title`: the bare
    centred page number, displayed only by the phone breakpoint while the
    chrome is hidden. */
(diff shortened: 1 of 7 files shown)


seamus-sloan (approved):
hell ye
image

PR #1489, opened 2026-07-28
feat: content validators for in-place book file and cover updates (#1231)

roberte777:
## Summary
- Closes #1231. Book files and covers can be replaced under a stable `books.uuid`, but nothing on the wire identified file *content* — so a device holding a downloaded copy served it forever, and a Range-based resume could splice the tail of a new file onto the head of an old one and store the result as a finished book.
- **Server** — every route that serves a file off disk now goes through `backend::validator`: it stamps an `ETag` derived from the *served artifact's* own stat, answers `If-None-Match` with a bodyless 304, and enforces `If-Range`. `/api/ebooks/{uuid}/file` also trades `max-age=86400` for `no-cache`, which is what makes a replaced book visible at all rather than a day later.
- **Clients** — both snapshot the validator at download time and compare it on the next metadata refresh to drive an "Update available" state with a re-download; the web engine sends `If-Range` on resume; both image caches revalidate a cached cover once per launch.
- **No migration.** `book_files.(mtime_epoch, size_bytes)` already exists and is already the indexer's own change-detection key, so the validator derives from data the reindex maintains.

Three things worth a reviewer's attention, since they are where the issue's plan and this implementation diverge:

>[!IMPORTANT]
> **`If-Range` could not be fixed from the client.** `tower_http::ServeFile` never looks at the header, so a client sending the precondition would still have been answered with a 206 counted from the new bytes. `enforce_if_range` strips the `Range` server-side when the tag no longer matches. This is the silent-corruption half of the issue and the reason the server change is a prerequisite, not a convenience.

>[!NOTE]
> **The cover half of the issue had already shipped.** `covers.rs` grew a content `ETag` + 304 in 3657d27b (the day after #1231 was filed, as a side effect of #1086), so AC4 was purely client-side. This PR reuses those helpers rather than adding a second recipe.

>[!NOTE]
> **The validator is resolved server-side.** `EbookMetadata.epub_validator` / `audio_validator` are hoisted the way `epub_size_bytes` already is, because `book_files` is only attached for multi-file books — a client-side pick would have had nothing to compare for the ordinary single-file book, and could disagree with the server about which file an unqualified download meant. `BookFileInfo.validator` stays for the explicit `?file_id=` case.

Emitted as a **strong** validator (nginx's `(mtime, size)` recipe), because `If-Range` is defined only under strong comparison — a `W/` prefix would make every resumed download restart from byte zero. The blind spot is a same-second, same-length replacement, which is the blind spot the indexer's own diff key already has.

## Test plan
- [x] `just lint` clean across the matrix (fmt, clippy incl. mobile + frontend-server + wasm32, stylelint).
- [x] `just test` — 3,380 Rust tests pass. New coverage: 13 in `backend::validator` (200/304/206, both `If-Range` directions, the date form, the unevaluable case), 3 in the web download engine (records the tag, sends `If-Range`, **discards its partial copy when the file changed under it**), 8 for `is_stale`, 4 in `db::books`, 5 in `shared::validator`, plus a route-level 304 test.
- [x] `xcodebuild test` — iOS suite passes, incl. 5 new tests for the wire shape and the per-kind pick.
- [x] **Live server**, real fixtures, not just the test harness: `ETag` present, `Cache-Control: private, no-cache`, 304 on match, 200 on a stale `If-None-Match`, 206 on plain and matching Range, and **200 — not a spliced 206 — on a stale `If-Range`**. This is also what caught the single-file `book_files` gap above.
- [ ] Playwright not run — no spec asserts on cache headers or the new testids, and the offline section is mobile-only (Maestro's territory per rule 04).
- [ ] Neither the mobile WebView nor the SwiftUI surface was driven visually; both compile and their logic is unit-tested.

## Version
By default a merge to `main` cuts a patch release. Pick the version update this
PR should trigger; labels override the default:
  (`vX.Y.Z` → `vX.Y.(Z+1)`).
- [ ] **No release** — add the `no release` label to skip cutting a release.

Left unlabeled, so the default patch applies. Criterion 2 could argue for minor — it completes a feature with new UX on both clients — so relabel if you read it that way. It is **not** a compatibility break: every new field is additive and `#[serde(default)]`, so an older mobile build ignores them and keeps working.

## Notes
- **Known gap, deliberate.** `/api/ebooks/{uuid}/download` serves the override-baked export, whose bytes change on a metadata edit while the source file's stat does not — so the web chip will not fire for that case. Resume correctness is unaffected there (that path uses the response `ETag`). Closing it needs a separate export validator; folding `books.last_modified` into `epub_validator` would over-signal on iOS, which downloads the source via `/file`. Flagged rather than scoped in silently.
- **Two pre-existing failures, unrelated to this branch.** `just dev-up` fails with *migration 49 was previously applied but has been modified* — the migration file is untouched in git, so it is local `omnibus.db` state; I used a scratch database rather than deleting anyone's. And `public_domain_epubs_parse_with_expected_metadata` fails with 22 fixtures on disk against 21 in `EXPECTED` — confirmed failing identically on unmodified `main` in a clean worktree.
- Docs: `docs/architecture.md` gains both new modules and the changed endpoint behaviour; `add-backend-route` gains the rule that a file route goes through `serve_download` / `validator::serve_file`, never `ServeFile` directly, with the reason.


frontend/src/offline/downloads/engine.rs
```diff
@@ -116,7 +116,13 @@ async fn run_inner(
         .unwrap_or_default();
     let mut files: Vec<PlannedFile> = Vec::with_capacity(plan.len());
     for mut f in plan {
-        if prior.iter().any(|p| p.rel == f.rel && p.done) {
+        let previous = prior.iter().find(|p| p.rel == f.rel);
+        // Carry the recorded validator across the replan, for finished files
+        // and half-finished ones alike: a fresh plan knows the URL but not
+        // what the bytes already on disk were served under, and dropping it
+        // would leave the resume with no precondition to send.
+        f.etag = previous.and_then(|p| p.etag.clone());
```
roberte777, line 183: [P1] Invalidate prior parts when the source validator changes. This merge trusts each prior `done` file based only on its path and existence, then the completed entry is stamped with the current book validator. If an audiobook changes between interrupted attempts, already-finished old tracks are retained while unfinished tracks download from the new source, producing a mixed set that is reported as current. Compare the prior entry-level validator with `source_validator` before reusing any completed files or partials, and discard the prior plan on mismatch.

frontend/src/offline/downloads/engine.rs
```diff
@@ -206,6 +237,17 @@ async fn download_file(
     let mut req = data::with_bearer(data::streaming_client().get(&url));
     if resumed > 0 {
         req = req.header(reqwest::header::RANGE, format!("bytes={resumed}-"));
+        // Without this the server cannot tell that the file was replaced
+        // since the interrupted attempt, and the 206 it returns would be
+        // counted from the *new* bytes — appending a new tail to the old head
+        // already in `.part` and leaving a corrupt book that nothing reports.
+        // A server that no longer recognises the tag answers with the whole
+        // file instead, which the 200 branch below restarts cleanly from.
+        if let Some(etag) = file.etag.as_deref() {
```
roberte777: [P2] Do not send a Range request when there is no usable validator. A `.part` can come from a legacy row, or survive a crash before the detached `persist` of its ETag completes. In that case this still sends `Range` without `If-Range`; the server can answer 206 and this code will append a new tail to an unknown head. If `resumed > 0` but no valid ETag is available, delete/truncate the partial and make a full request. Ideally, also make the ETag durable before writing response bytes.

frontend/src/pages/book_detail/offline.rs
```diff
@@ -108,6 +113,21 @@ fn DlRow(
         let uuid = uuid.clone();
         move |_| downloads::remove(&uuid, format)
     };
+    // Remove first: `start` refuses to run against a Complete entry, and the
+    // bytes on disk are the superseded ones either way.
+    let redownload = {
+        let (uuid, title, server_url) = (uuid.clone(), title.clone(), server_url.clone());
+        move |_| {
+            downloads::remove(&uuid, format);
```
roberte777: [P1] This races the asynchronous cleanup started by `remove`. The registry is cleared synchronously, but the database deletion and filesystem removal happen later on the runtime; `start` immediately inserts the replacement and begins writing into the same directory. The delayed cleanup can therefore delete the new row or its new files/`.part`. Please make re-download a single replacement operation that stages new bytes and swaps them after success (or, at minimum, await cleanup before starting).

roberte777 (commented):
The underlying problem is valid, and the server-side validator/ETag/304/If-Range work is the right foundation. I would keep that, along with the additive metadata fields and cache revalidation. I found four correctness issues in the client replacement/resume paths that should be addressed before merge; one applies to both clients, so I left five inline anchors. The main risks are corrupt mixed-version audiobook resumes, unconditioned legacy partial resumes, a web cleanup/start race, deleting the only usable offline copy before replacement succeeds, and recording an iOS download under a validator that may not describe the response bytes.

omnibus-ios/omnibus/Offline/DownloadManager.swift
```diff
@@ -110,6 +110,30 @@ final class DownloadManager: NSObject {
...
+        guard let record = record(for: book.uuid, kind: kind), record.state == .complete,
+              let takenUnder = record.validator,
+              let current = book.sourceValidator(for: kind)
+        else { return false }
+        return current != takenUnder
+    }
+
+    /// Replace a superseded copy with the file the server holds now.
+    func redownload(_ book: Book, kind: DownloadKind) async {
+        await remove(book.uuid, kind: kind)
```
roberte777: [P1] Re-download should preserve the usable old copy until the replacement succeeds. Removing first means a transient network, auth, storage, or server failure leaves the reader with no offline book. Download to a separate staged destination and atomically replace the completed file/record only after the new transfer is verified; otherwise I would defer the Re-download action and keep only the stale warning.

omnibus-ios/omnibus/Offline/DownloadManager.swift
```diff
@@ -146,7 +170,10 @@ final class DownloadManager: NSObject {
         let record = DownloadRecord(
             bookUUID: uuid, kind: kind, format: format, state: .running, localPath: nil,
             totalBytes: 0, receivedBytes: 0, updatedAt: Int64(Date().timeIntervalSince1970),
-            error: nil
+            error: nil,
+            // Snapshotted from the book being downloaded, so a later refresh
+            // can tell this copy from the one the server holds.
+            validator: book.sourceValidator(for: kind)
```
roberte777, line 213: [P2] Bind this validator to the actual HTTP response, not to metadata observed before the transfer. If the source changes between the metadata fetch and the download response, the bytes that finish here can be recorded under the wrong validator and immediately appear stale (or fail to be recognized as stale under the opposite timing). Have the download response carry the source/content validator and persist that value from `HTTPURLResponse` when marking the record complete.

frontend/src/offline/downloads/engine.rs
roberte777: Fixed in 66a3fa4. Confirmed the scenario: the merge trusted `done` files on path + existence alone and the entry was then stamped with the current source validator, so a multi-part audiobook could hold old tracks and new ones and report itself current.

`run_inner` now pulls the whole prior entry rather than just its `files`, and only reuses it when `entry.validator` still equals the freshly computed `source_validator`. Unprovable counts as unusable — an entry from before validators were recorded is discarded too, costing one restart. Discarding also disarms every resume below it, since the carried `etag`s go with the plan and `download_file` refuses a partial it cannot vouch for.

Worth saying plainly: this was the same bug the PR claims to fix, one level up. The body said "never a spliced mix" while leaving file-granularity mixing open.

frontend/src/offline/downloads/engine.rs
roberte777: Fixed in 66a3fa4, and you were right that this is worse than P2 — I would call it P1. The crash window is narrow, but the legacy path is not: every `.part` that exists at upgrade time deserializes with `etag: None`, so the first resume after this ships was unguarded and could splice silently.

`download_file` now emits no `Range` at all without a precondition to qualify it; the request falls through to the existing restart branch, which truncates the partial and re-counts. Regression test `download_file_does_not_resume_a_partial_it_has_no_validator_for` asserts the outbound `Range` header is empty and the stale bytes are dropped.

Also took the "ideally" — `set_file_etag` is now awaited via `downloads_upsert` instead of `run_detached`, so the validator is on disk before the first body byte.

frontend/src/pages/book_detail/offline.rs
roberte777: Fixed in 66a3fa4. Confirmed, and there is a sharper failure than the race: `start` fast-fails when offline, so tapping "Re-download" with no connection deleted the book and then refused to fetch it. Guaranteed data loss, no interleaving required.

Replaced remove-then-start with `downloads::replace`, which never touches the live copy. The transfer stages under `{uuid}/.replacing/` and `swap_into_place` renames each file over the live one only once every file has landed; a failure removes the staging dir and restores the prior entry verbatim, so the reader keeps a readable book and the staleness chip stays as the retry affordance. The control is also disabled while offline, per rule 08 corollary.

One thing your comment did not mention that I fixed alongside: the old handler passed `file_id: None`, so replacing a specific file on a multi-file book would have fetched a different one. `replace` reuses the prior entry file_id.

omnibus-ios/omnibus/Offline/DownloadManager.swift
roberte777: Fixed in 66a3fa4. Agreed — and this applies to the web client too, which your other comment framed only as a race; the same remove-then-start destroys the copy there. Treated them as one design.

iOS already stages (the transfer lands in a temp file and `didFinishDownloadingTo` only swaps on a 2xx), so the fix was to stop deleting first. `redownload` records the superseded record in `superseding`, `start` keeps its `localPath` so the bytes stay addressable, and all three failure paths — non-2xx, failed move, transport error — restore it instead of writing a `.failed` row over a file that is still perfectly readable. Cancels clear the slot without restoring. The control is disabled while offline.

omnibus-ios/omnibus/Offline/DownloadManager.swift
roberte777: Fixed in 66a3fa4, though not the way the comment literally reads, and I want to be explicit about why.

Persisting the `ETag` alone would have been wrong for multi-part audiobooks: `/api/audiobooks/{uuid}/download` serves the lowest-ordinal *part*, so its `ETag` is that part stat, while `audio_validator` is the `book_files` group aggregate. Those coincide for a single-part m4b and diverge for anything multi-part, which would have reported that book as permanently updatable with re-downloading never clearing it — worse than the race being fixed.

So I took the first half of your remedy: the response now carries the source validator explicitly, as `X-Omnibus-Source-Validator`, naming the `book_files` row it drew from. Both clients persist that and fall back to the pre-transfer metadata read when it is absent. Server side it is `file_etag(source)` on the ebook routes (where the source path is already in hand) and a new `db::book_file_validator(book_file_id)` on the audiobook routes. Verified against a live server that the header equals the metadata validator on both `/file` and `/download`, which the `ETag` does not.

shared/src/validator/tests.rs
```diff
@@ -0,0 +1,36 @@
+use super::*;
+
```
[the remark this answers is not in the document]
roberte777: Fixed in 66a3fa4 — added the module doc.

shared/src/validator.rs
```diff
@@ -0,0 +1,36 @@
+//! The content validator (HTTP `ETag`) for a file served off disk.
+//!
+//! One recipe, shared by the db layer — which reports a file's current
+//! validator alongside its metadata — and the server handlers, which stamp
+//! the same string on the response. A client compares the two to decide
+//! whether the copy it downloaded is still the copy the server holds, so the
+//! two sides producing byte-identical strings is the whole contract.
```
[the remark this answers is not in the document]
roberte777: Fixed in 66a3fa4 — trimmed to 5 lines.

server/src/backend/validator.rs
```diff
@@ -0,0 +1,148 @@
+//! `ETag` stamping and the conditional-request preconditions that depend on
+//! it, for the routes that serve a file off disk.
+//!
+//! `ServeFile` handles `Range` but knows nothing about entity tags, so the
+//! handlers evaluate `If-None-Match` and `If-Range` here before delegating to
+//! it and stamp the validator on the way back out.
```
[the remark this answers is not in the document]
roberte777: Fixed in 66a3fa4 — trimmed to 4 lines.

PR #1492, opened 2026-07-29
feat: content validators and conditional-request handling

roberte777:
First of six PRs for #1231. Establishes the content validator and applies it to the ebook byte endpoints.

## Why

Book files can be replaced in place under a stable `books.uuid` — a re-uploaded EPUB, a corrected file rsynced back — but nothing on the wire identified file *content*. A client holding a downloaded copy had no way to learn its copy had gone stale, and a `Range`-based resume could splice the tail of a new file onto the head of an old one.

## What

**`omnibus_shared::file_etag`** derives a validator from the `(mtime_epoch, size_bytes)` pair `book_files` already records. That pair *is* the reindex diff key (`db/src/indexer/mod.rs`), so "the validator moved" and "the indexer classified this Changed" are one statement rather than two that can drift.

No new column, and so **no migration**. The issue proposed storing a validator; a stored column would be a second copy of a fact that every insert/backfill site must remember to bump — the exact drift the issue says it wants to avoid. Deriving it cannot drift.

**`backend::conditional`** answers `If-None-Match` with a 304 and — the reason it exists — drops a `Range` whose `If-Range` no longer matches, so the file service returns the whole new body instead of a mismatched 206. `tower-http` 0.7 already emits an ETag and handles `If-None-Match`, but has no `If-Range` support at all; that gap is the corruption bug.

Its request-time validator is deliberately **stronger** than the wire one: `(inode, mtime.nanos, size)`. A stat-derived tag's real weakness is a replacement that preserves the timestamp — `rsync -t`, `cp -p`, a restore from backup — and almost every such tool writes a temp file and renames it, which moves the inode even when the mtime comes along unchanged. This is Apache's historic `FileETag INode MTime Size`; it dropped INode only because it breaks multi-server clusters, which a self-hosted single instance is not.

The two validators answer different questions and that is why they differ:

| | Source | Granularity | Answers |
|---|---|---|---|
| Response ETag | the served file's stat, at request time | `(ino, secs.nanos, size)` | `If-None-Match`, `If-Range` |
| Wire etag (PR 2) | `book_files` columns | `(mtime_epoch, size)`, seconds | "is my download stale?" |

A miss on the first splices a file. A miss on the second merely delays a notification, and it cannot be stronger than the diff key it mirrors anyway.

**`/api/ebooks/{uuid}/file`** now streams via `ServeFile` instead of buffering the whole book into memory, which is what gives it `Range` support — without that, the `If-Range` guarantee would have nothing to guard.

## Behaviour changes worth a second look

- `/file` and `/download` gain `Cache-Control: private, no-cache`. `/file` previously had `max-age=86400`, which meant a browser holding a day-old copy of a replaced book never even asked. Revalidation now costs one 304.
- `/file`'s `Vary` widens from `Cookie` to `Cookie, Authorization`. The route authenticates by cookie *or* bearer *or* `?token=`; now that it carries an ETag, a shared cache keying only on `Cookie` could hand one user's 304 — hence their copy of the book — to another. Same reasoning the cover handler already documents.
- `/file` returning a missing file as 404 is preserved explicitly rather than inherited from `ServeFile`.

## Residual, stated plainly

An explicit in-place overwrite that preserves both length and timestamp (`rsync --inplace`, `dd`) still slips through. No stat-derived validator catches that, and a content hash would mean re-reading every audiobook on every scan. PR 3 closes it from the other end, by verifying the assembled file parses rather than trusting the validator alone.

## Tests

- 21 unit tests on `conditional`: every `If-None-Match` form (exact, weak, `*`, list, stale, no-validator) and every `If-Range` arm (match, stale, weak, date, no-validator, bare `Range`), plus that consumed conditionals are stripped before delegation.
- `etag_for_path_changes_when_the_file_is_replaced` performs the real attack: same length, mtime copied forward, replaced by rename. Only the inode term distinguishes them.
- 6 endpoint tests covering 200-carries-ETag, 304 revalidation, a matching resume returning 206 with the right slice, and — the one that matters — a stale resume returning **200 with the whole body**.

`just lint` and `just test` pass. (`omnibus-db::public_domain_epubs` fails on my machine from a stray local file in the gitignored fixtures dir — 22 EPUBs present, 21 expected — unrelated to this diff, which touches only `shared/` and `server/`.)

## Stack

1. **feat/content-validators** ← this PR
2. audiobook endpoints + `BookFileInfo.etag` wire field + rule 09
3. offline engine: snapshot, `If-Range` on resume, container verify
4. "Update available" UI
5. offline image cache revalidation
6. iOS

Refs #1231


server/src/backend/conditional.rs
```diff
@@ -0,0 +1,196 @@
...
+            .is_some_and(|value| if_none_match_hits(value, etag))
+        {
+            return Precondition::NotModified(etag.to_string());
+        }
+    }
+    let stale_range = header_str(req.headers(), header::IF_RANGE)
+        .is_some_and(|value| !etag.is_some_and(|etag| if_range_matches(value, etag)));
+
+    let headers = req.headers_mut();
+    headers.remove(header::IF_NONE_MATCH);
```
roberte777: A stale `If-None-Match` can still produce a false 304 here. For example, if the request also carries an `If-Modified-Since` date matching the current file, this function removes the stale ETag condition but leaves the date condition for `ServeFile`; `ServeFile` then returns 304 even though RFC precedence says the non-matching `If-None-Match` requires the full body. `If-Match` and `If-Unmodified-Since` are likewise evaluated downstream against the different ETag emitted by `ServeFile`. Please make this layer evaluate the complete precondition precedence against the canonical validator/mtime and strip every condition it settles before delegation.

server/src/backend/ebooks.rs
```diff
@@ -252,30 +260,75 @@ pub(super) async fn get_ebook_file(
...
         }
+    };
+    let (mut parts, body) = res.into_parts();
+    if !matches!(
+        parts.status,
+        StatusCode::OK | StatusCode::PARTIAL_CONTENT | StatusCode::NOT_MODIFIED
+    ) {
+        // A missing file reaches here as ServeFile's own 404; keep the
+        // endpoint's historical contract of reporting that as 404.
+        return StatusCode::NOT_FOUND.into_response();
```
roberte777: Please preserve non-404 statuses from `ServeFile`. An unsatisfiable or malformed range legitimately returns 416 with `Content-Range: bytes */<length>`; converting it to 404 makes resume clients think the book disappeared and discards the information they need to restart correctly. Only translate an actual missing-file response, and pass statuses such as 416 through unchanged.

server/src/backend/ebooks.rs
```diff
@@ -252,30 +260,75 @@ pub(super) async fn get_ebook_file(
...
-            [
-                (header::CONTENT_TYPE, "application/epub+zip"),
-                (header::CACHE_CONTROL, "private, max-age=86400"),
-                (header::VARY, "Cookie"),
-                (header::X_CONTENT_TYPE_OPTIONS, "nosniff"),
-            ],
-            bytes,
-        )
-            .into_response(),
+    let etag = conditional::etag_for_path(&path).await;
```
roberte777: The validator is derived from pathname metadata before `ServeFile` opens that pathname. If an atomic replacement occurs between this line and `ServeFile::oneshot`, a request whose `If-Range` matches the old ETag keeps its range while `ServeFile` opens the new bytes—the splice this code promises to prevent. To make the integrity guarantee hold, derive the validator and stream the representation from the same opened file handle, or otherwise detect that the path changed and restart with a full response.

PR #1494, opened 2026-07-29
feat: publish content validators on the media endpoints

roberte777:
Second of six for #1231. **Stacked on #1492** — review that first; this PR's diff is against it.

Completes the server half of AC1/AC2 and puts the wire validator clients need for AC3 on `BookFileInfo`.

## What

**Audiobook endpoints.** `/api/audiobooks/{uuid}/parts/{ordinal}` and `/api/audiobooks/{uuid}/download` now carry an `ETag`, answer `If-None-Match` with a 304, and drop a `Range` whose `If-Range` went stale.

Each part validates against its **own** file rather than the `book_files` row, which holds the group's `max(mtime)` and summed size across every part — the wrong granularity for a per-part URL. Two parts of the same book get two validators, as they should.

**`BookFileInfo.etag`** carries `file_etag(size_bytes, mtime_epoch)` from the `book_files` projection. A client snapshots it at download time and compares it against a later metadata refresh — no extra request, because that refresh already happens (the mobile data layer is network-first).

`None` for rows the scanner has not stat'd. `(0, 0)` is the indexer's never-observed sentinel, and publishing a validator for it would make the one-time stat backfill look like a content change on every device holding a download.

**Small consolidation.** `MEDIA_CACHE_CONTROL` and `with_media_cache_policy` move into `conditional` now that covers, ebooks and audiobooks all want them — so the 200 and 304 paths cannot drift on `ETag`/`Cache-Control`/`Vary`.

## Docs

New [`.claude/rules/09-content-validators.md`](.claude/rules/09-content-validators.md) records the invariants that keep this honest, because they are the kind that erode quietly:

- The validator is **derived from the reindex diff key**, never stored and never a counter, so "the validator moved" and "the indexer classified this Changed" cannot disagree.
- The response validator and the wire validator are deliberately different strengths and must not be collapsed into one.
- Both are **strong** tags — RFC 9110 §13.1.5 requires strong comparison for `If-Range`, so a weak tag would make every resume restart from zero.
- `ETag` and `Vary` travel together on these auth-varying routes.
- Every new byte-serving endpoint routes through `conditional::apply`; `ServeFile` alone is not enough, because tower-http has no `If-Range` support.

Also updated: the CLAUDE.md rules index, both module maps in `docs/architecture.md`, and a rule-98 pass adding the validator step to the `add-backend-route` skill.

## Tests

- 6 audiobook endpoint tests: 200 carries an `ETag` and the right `Vary`; 304 revalidation; a matching resume returns 206 with the correct slice; a **stale resume returns 200 with the whole body**; the validator moves when a part is replaced.
- 2 db projection tests: the validator is the `(mtime_epoch, size_bytes)` pair, and the never-observed sentinel projects `None`.
- The pre-existing bare-`Range` test still passes — an unconditional `Range` is a request for a slice, not a consistency guarantee, so audiobook seeking is untouched.

## A gap I hit, worth knowing

`just lint` clippies `-p omnibus-frontend --features server` and `-p omnibus-mobile`, but not `-p omnibus-frontend --features mobile --all-targets` — so a test-only constructor under the `mobile` feature compiled clean through lint and only failed under `just test`.

Separately, `just test`'s `&&` chain means the first crate's failure hides every later crate. On my machine `omnibus-db::public_domain_epubs` fails from a stray file in the gitignored fixtures dir (22 EPUBs present, 21 expected), which masked the rest until I ran each crate individually. Neither is in scope here; flagging both.

Full matrix run per-crate: db 1476 passed (1 skipped), server 584, frontend/server 533, frontend/mobile 556, shared 238. `just lint` clean.

## Stack

1. [#1492](https://github.com/seamus-sloan/omnibus/pull/1492) validators + conditional handling + ebook endpoints
2. **feat/publish-content-validators** ← this PR
3. offline engine: snapshot, `If-Range` on resume, container verify
4. "Update available" UI
5. offline image cache revalidation
6. iOS

Refs #1231


roberte777 (commented):
Review scoped only to changes introduced in #1494.

db/src/books/get.rs
```diff
@@ -489,27 +489,34 @@ where
...
+                    format,
+                    filename,
+                    ordinal,
+                    label,
+                    size_bytes,
+                    path,
+                    // Same `(mtime_epoch, size_bytes)` pair the reindex diff
+                    // keys on, so a file the indexer classified Changed is
+                    // exactly a file whose validator moved.
+                    etag: omnibus_shared::file_etag(size_bytes, mtime_epoch),
```
roberte777, line 517: The new `BookFileInfo.etag` is derived here, but it will not reach the common single-file API response: `get_book` only assigns `book.book_files` when one format has multiple rows. A normal one-EPUB or one-M4B book therefore still serializes no `BookFileInfo`, leaving clients with no validator to snapshot or compare. Please expose the targeted file validator for single-file books too, and add an API/serialization test for that case.

server/src/backend/audiobooks.rs
```diff
@@ -236,6 +237,18 @@ pub(super) async fn get_audiobook_part(
 
     let abs_path = std::path::Path::new(&resolved.library_path).join(&part.filename);
     let mime = audiobook::mime_for_filename(&part.filename);
+
+    // The validator comes from this part's own file, not the `book_files`
+    // row: that row carries the group's `max(mtime)` and summed size across
+    // every part, which is the wrong granularity for a per-part URL.
+    let etag = conditional::etag_for_path(&abs_path).await;
```
roberte777: This new audiobook path stats the pathname for its ETag, then `ServeFile` opens that pathname separately. If the file is atomically replaced between those operations, an `If-Range` matching the old ETag keeps its `Range` while `ServeFile` opens the new bytes, producing the mixed-file 206 this change is meant to prevent. Please bind validation and streaming to the same opened file identity, or detect a changed identity after opening and restart with a full 200. The new download path below has the same race.

server/src/backend/audiobooks.rs
```diff
@@ -247,7 +260,21 @@ pub(super) async fn get_audiobook_part(
             .headers
             .insert(axum::http::header::CONTENT_TYPE, value);
     }
-    Response::from_parts(parts_resp, Body::new(body))
+    // An `ETag` on a route that authenticates by cookie *or* bearer *or*
+    // `?token=` needs a `Vary` naming all of them, or a shared cache can
+    // hand one user's 304 to another.
+    for (name, value) in [
```
roberte777: This new cache-policy loop runs for every `ServeFile` status, so 404 and 416 responses also receive the media `Cache-Control`/`Vary` headers. Please gate these headers on 200/206, ideally by passing the response through `conditional::with_media_cache_policy` as the download path does.

PR #1610, opened 2026-08-02
fix: keep HardcoverFetchPanel hook order stable across availability transition (#1590)

roberte777:
## Summary
- Closes #1590
- `HardcoverFetchPanel` declared `state`/`busy` via `use_signal` *after* the `if !available() { return rsx! {}; }` early return, so the component called a different number of hooks before vs. after the async availability check resolved on the same mounted instance — a hook-order violation of rule 07 (`.claude/rules/07-hydration.md`), not merely SSR/hydration parity.
- Fix: move both `use_signal` declarations above the `available` gate, exactly per the issue's suggested approach. Hook count/order is now identical on every render; the gate still short-circuits only the visible markup.
- Added a regression test (`frontend/src/pages/metadata_edit/hardcover_fetch/tests.rs`) that mounts a harness mirroring the fixed hook shape in a real `VirtualDom`, publishes its `available` `Signal<bool>` back to the test via `use_hook`, flips it `false -> true` through `dom.in_runtime(|| available.set(true))` (the same signal-write scheduling path production uses), forces a second render with `render_immediate`, and asserts the *rendered content* is correct post-flip (not just that nothing panicked).

## Note on test scope
The real `HardcoverFetchPanel`'s availability check goes through a `dioxus::fullstack` server function whose extractors (`pool: PoolExt`, `user: AuthUser`) need a live axum request context to resolve — there's no seam in this crate's existing test harness to drive that resolution from a bare `VirtualDom` (unlike `rpc::overrides`'s tests, which call an *extracted* pool-taking body directly; this rpc is a one-line lookup with nothing to extract). The added test instead mirrors the panel's exact hook-declaration shape and drives the identical false→true transition with a real `Signal<bool>`, which exercises the actual Dioxus hook-slot invariant rule 07 requires.

Also worth noting: this build of Dioxus (v0.7.9) does not hard-panic on this specific *monotonic-growth* hook-count change (a render calling strictly more hooks than the previous one just extends the hook list) — confirmed by running the pre-fix shape through the same harness during development, which completed without panicking. That's consistent with the issue's own wording ("risking a panic or silently misattributed signal state") — the invariant in rule 07 exists independent of whether this exact transition crashes on this exact Dioxus build, and the added test guards the *content* correctness of the fixed shape, not just crash-freedom.

## Test plan
- [x] `cargo fmt -p omnibus-frontend` — clean
- [x] `cargo clippy -p omnibus-frontend --all-targets --features server -- -D warnings` — PASS, no warnings
- [x] `cargo test -p omnibus-frontend --features server` — PASS, 618 passed; 0 failed

## Version
- [x] **Patch** — left unlabeled (default)

## Notes
- Manual verification via `ui-validate` was not run for this change (the fix is a hook-declaration reorder with identical visible behavior); the new unit test exercises the exact false→true transition the issue describes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DhPxqqQHT7EfpDyhQGZHWx)_

### DIFF
--- a/frontend/src/pages/metadata_edit/hardcover_fetch.rs
+++ b/frontend/src/pages/metadata_edit/hardcover_fetch.rs
@@ -17,6 +17,9 @@ use omnibus_shared::metadata_fetch::{FetchedBookMetadata, HardcoverFetchResult};
 use super::form_grid::FormFields;
 use crate::{data, use_server_url};
 
+#[cfg(test)]
+mod tests;
+
 /// What the fetch action is currently doing.
 #[derive(Clone, PartialEq)]
 enum FetchState {
@@ -34,6 +37,13 @@ enum FetchState {
 pub(super) fn HardcoverFetchPanel(uuid: String, fields: FormFields) -> Element {
     let server_url = use_server_url();
     let mut available = use_signal(|| false);
+    // Declared unconditionally, before the `available` gate below, so this
+    // scope calls the same number of hooks in the same order on every
+    // render — the gate only skips the visible markup, never a hook
+    // (rule 07; #1590 was this component calling one hook pre-availability
+    // and three post-availability on the same mounted instance).
+    let mut state = use_signal(|| FetchState::Idle);
+    let mut busy = use_signal(|| false);
 
     let url = server_url.clone();
     use_effect(move || {
@@ -49,8 +59,6 @@ pub(super) fn HardcoverFetchPanel(uuid: String, fields: FormFields) -> Element {
         return rsx! {};
     }
 
-    let mut state = use_signal(|| FetchState::Idle);
-    let mut busy = use_signal(|| false);
     let on_fetch = move |_| {
         if busy() {
             return;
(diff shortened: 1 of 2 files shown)


roberte777:
Closing as superseded — not because the work was wrong, but because it already landed by another route.

[#1670](https://github.com/seamus-sloan/omnibus/pull/1670) (commit `21286a03`, *"fix: declare HardcoverFetchPanel hooks unconditionally before the availability gate (#1590)"*) is on `main` and carries both halves of this PR:

- **The production fix** — `state`/`busy` are declared above the `if !available()` gate in `frontend/src/pages/metadata_edit/hardcover_fetch.rs`, exactly as here.
- **The regression test** — an inline `#[cfg(all(test, feature = "server"))] mod tests` in that same file, using the same structural-mirror gate harness this PR puts in a sibling `hardcover_fetch/tests.rs`.

Issue #1590 is closed. Rebasing this branch would re-apply a fix `main` already has and move a test file for no behavioural gain, so there is nothing left to merge.

PR #2219, opened 2026-08-27
feat: make the Continue widget a one-book call to action

roberte777:
## Summary
- Rebuilds the Continue widget's `systemSmall` and `systemMedium` families as a one-book call to action; `systemLarge` is untouched and still lists five.
- Medium becomes the app's `ContinueHero` translated to a widget: the cover at full card height, a `CONTINUE READING` / `CONTINUE LISTENING` eyebrow, the title at 17pt serif instead of 11.5pt, author, progress bar, position readout, and a Read/Play pill that resumes the book straight from the Home Screen. It replaces the three-across fan, which showed three thumbnails and three truncated titles with nothing on the card to act on.
- Small draws the book's own artwork blurred and saturated into its ground, the sharp cover on top under a graded scrim, and runs the progress bar edge to edge along the bottom of the card.
- A widget cannot be swiped — the system renders it as a snapshot and only routes taps — so the rest of the rail is reached by a chevron backed by `ShowNextBook`, an `AppIntent` that moves a cursor kept in the App Group's defaults and reloads the timeline. Dots beside the eyebrow show where you are.
- The flip rail caps at three (`WidgetSnapshot.flipRail`). Five is what the large family *lists*, and listing is cheap; every book past the front of a flip costs a tap and a redraw to reach, and past three the control stops meaning "the other book I'm in" and starts meaning "page my library".
- Small carries no flip control at all. It is 130pt square on a 6.3" phone, a chevron there spends the card's only spare corner arguing with the artwork, and "what am I in the middle of" is answered by whichever book you touched last.
- The cursor names a **book** (`WidgetBook.id`), never an index. The rail is rewritten on every position save, so an index would mean "whatever is second today" and the next save that reordered it would move the card out from under the reader. A cursor whose book has left the rail — finished, or pushed past the three-book cap by a newer one — falls back to the front rather than resolving to a stale position.

## Test plan
- [x] `just ios-build` — app plus widget extension compile clean.
- [x] `just ios-test` — 544 tests pass, including 11 new `WidgetRailCursorTests` covering cursor resolution, reorder stability, the fall-back when a book leaves the rail, wrap-around, the three-book cap, and the dual-format case where `a:epub` and `a:audio` must not resolve to each other.
- [ ] **Not verified on a real Home Screen.** Every screenshot below is an offline render harness that compiles the extension's view sources for the simulator and draws the cards; it does not run the widget process. So the chevron is drawn but never tapped, and `ShowNextBook`'s round trip (cursor write → timeline reload → card advances) is covered by unit tests on `next(after:)` and by nothing else. `containerBackground` is supplied by the harness rather than by WidgetKit. Worth adding the widget by hand via `just ios-sim` before or after merge.

## Version
- [x] **Patch** — the default; leave unlabeled.

## Notes
- Layout numbers on both cards are budgets rather than preferences, and the binding constraint is the 6.3" phone: `systemSmall` is 130pt of content there, and the medium card's dial and title compete for the same 130. Both families lost the title's second line to an oversized element at least once during this work, and the regression is invisible on a 6.9" phone — which is why the render sheet poses every card at both point sizes and in both colour schemes.
- The middle commit builds a Smart Stack approach (`AppIntentConfiguration`, a per-book `SelectBookIntent`, `relevance()`) and the third removes it. Reaching the rail that way meant hand-pinning one widget per book and stacking them, which is setup most readers will never do; the chevron does it in one tap with none. Squash-merge hides the round trip, but it is in the commit list.
- Known rough edges, deliberately left: long titles still truncate on medium at 17pt serif, and `systemLarge` is still the five-row list rather than being reworked to match the one-book direction.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -399,10 +399,9 @@ are the other half of a second target:
 OmnibusShared/      — source compiled into *both* the app and the widget
                       extension (a filesystem-synchronized group listed in two
                       targets' `fileSystemSynchronizedGroups`): WidgetSnapshot
-                      (the wire form, plus the cursor arithmetic the one-book
-                      families walk their rail with), WidgetStore (the App
-                      Group container's layout, the only encoder either side
-                      uses, and where that cursor is kept), DeepLink
+                      (the wire form, plus which book a pinned widget resolves
+                      to), WidgetStore (the App Group container's layout, and
+                      the only encoder either side uses), DeepLink
                       (the `omnibus://book/<uuid>` URLs the widget mints and
                       the app parses back) and OKLCH (the OKLCH↔sRGB
                       conversion for server-provided accents — shared so the
@@ -414,8 +413,11 @@ OmnibusWidgets/     — the WidgetKit extension (`com.omnibus.mobile.widgets`,
                       else, so it renders in Airplane Mode with the app force
                       quit), its three family layouts and the three distinct
                       empty states, ContinueIntents (a widget cannot be swiped,
-                      so small and medium each show one book and carry an
-                      `AppIntent` that flips to the next), and WidgetTheme —
+                      so small and medium each show *one* book and the widget
+                      is an `AppIntentConfiguration` — one copy per book,
+                      stacked, and the swipe is the stack's; `relevance()`
+                      publishes one configuration per book in progress so a
+                      Smart Stack has something to rotate), and WidgetTheme —
                       colours derived entirely from the book's own tone,
                       because the reader's chosen
                       theme lives in the app's `UserDefaults`, which the
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -399,9 +399,10 @@ are the other half of a second target:
 OmnibusShared/      — source compiled into *both* the app and the widget
                       extension (a filesystem-synchronized group listed in two
                       targets' `fileSystemSynchronizedGroups`): WidgetSnapshot
-                      (the wire form, plus which book a pinned widget resolves
-                      to), WidgetStore (the App Group container's layout, and
-                      the only encoder either side uses), DeepLink
+                      (the wire form, plus the cursor arithmetic the medium
+                      card flips its rail with), WidgetStore (the App Group
+                      container's layout, the only encoder either side uses,
+                      and where that cursor is kept), DeepLink
                       (the `omnibus://book/<uuid>` URLs the widget mints and
                       the app parses back) and OKLCH (the OKLCH↔sRGB
                       conversion for server-provided accents — shared so the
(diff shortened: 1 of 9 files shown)


roberte777:
Round 1:
- Fixed: `WidgetArt` claimed in its doc comment to keep the small hero to one decode per file but had no memo — added one, keyed on `(name, modification date)` so a cover rewritten in place under the same name isn't served stale from a live extension process.
- Fixed: the hero backdrop blurred with `opaque: true`, which pulls black in around transparent edges — on exactly the covers `WidgetStore.thumbExtensions` files as PNG to keep unflattened. Now `opaque: false`.
- Fixed: the bloom offset's `43 - 85` were hand-derived from the cover width and bloom diameter; both are named `Layout` constants now and the offset is computed from them.
- Dismissed: nothing this round.
- Note: the multi-lens workflow review did **not** run — all four review agents died on a session limit, so the empty result is an absence of review rather than a clean one. It has to be re-run before this is merged.

roberte777:
Round 2:
- Fixed: `WidgetArt.memo` was not bounded despite its comment saying so — keying on `(name, modification date)` left the superseded entry behind on every cover rewrite. Now keyed by name with the date carried for comparison, and the map is dropped wholesale at `WidgetSnapshot.maxBooks`.
- Fixed: the flip chevron's hit region was its 22pt frame — half Apple's minimum, and the one control on the card whose miss does something *other* than what a hit does, since the surrounding area is the card-wide `widgetURL`. Tappable area padded out to 38pt with the layout footprint held at 22pt, so the eyebrow row still costs the title nothing.
- Fixed: `ResumePill`'s SF Symbol announced as "play.fill" to VoiceOver; hidden from accessibility, matching how `LargeRow` already handles the same thing.
- Fixed: `omnibus-ios/README.md` still described the medium family as "three across" — the layout this branch deletes — and documented the extension as read-only against the shared container, which the cursor write is not. (Found independently by three review lenses; `docs/architecture.md` had been updated but its sibling had not.)
- Dismissed: nothing this round.

roberte777:
Round 3:
- Fixed: the hero backdrop's veil was attached to the aspect-fill image, which reports a size half again taller than the square card — so the gradient graded that taller box and the card showed only its middle two thirds (0.22 of black at the top where 0.12 was specified, 0.62 at the bottom where 0.72 was). Now held in a `Color.clear` container that cannot grow, matching `WidgetCover` and the pattern the iOS README prescribes.
- Fixed: the flip control moved to an overlay. Copilot was right that `.padding(-slop)` also shrinks the bounds used for hit-testing, so the target stayed 22pt — but its suggested fix (accept the larger footprint) costs the title its second line at 158pt, which the render sheet confirms. An overlay buys the 40pt target without spending layout, and the eyebrow row gives 10pt back.
- Fixed: `ContinueWidget.swift` and `WidgetSnapshot.swift` headers still said the extension reads only the snapshot — the same claim corrected in the README last round.
- Dismissed: nothing this round.

roberte777:
Round 4:
- Fixed: moving the flip control into an overlay collapsed the eyebrow row to the height of its 9.5pt text, which pulled the title up under the control — the end of a long first line landed inside the chevron's tap circle, so tapping a word of the title flipped the book instead of opening it (8 of 14 sampled two-line titles reached it). The row holds 22pt open again, and the target grows up and outwards into the margin rather than down into the title.
- Dismissed: the small card's bar loses the leading corner of its fill to the widget's ~21.5pt mask, so a book under ~7% draws nothing. Real, but insetting the bar to clear the mask was tried and looked worse — a bar floating above the edge is a worse card than one whose first few percent don't show. Recorded at the call site as accepted rather than unknown, so it doesn't get re-raised as a discovery.

